PR #28, opened 2021-11-24
RateController

olofkallander:
Measures uplink using RTCP receiver reports and probing from sender side.
Video is paced according to estimate. If we do not do that, the burst from the sender will cause packet drop on the downlink at the receiver. Pacing prevents NACK on video that cause unnecessary additional load on the limited link. 
A positive effect of this is that if link is overloaded by video, the audio will still pass through as it is not paced. For example, video and screen share may be delayed and broken but audio will likely be intelligable. 

A limit is imposed on slides video stream according to estiamte but with a configurable min bandwidth. This means that we can set a minimum quality to allow users in meetings to have at least that level even if someone in the meeting does not have enough bandwidth. The person with insufficient bandwidth will have poor video but good audio. Video will be delayed and it will take time before a new slide can be seen but he can still participate in the meeting.

### DIFF
--- a/bridge/ApiRequestHandler.cpp
+++ b/bridge/ApiRequestHandler.cpp
@@ -381,6 +381,11 @@ ApiRequestHandler::ApiRequestHandler(bridge::MixerManager& mixerManager, transpo
       _legacyApiRequestHandler(std::make_unique<LegacyApiRequestHandler>(mixerManager, sslDtls))
 #endif
 {
+#if ENABLE_LEGACY_API
+    logger::logAlways("Legacy colibri API enabled", "ApiRequestHandler");
+#else
+    logger::logAlways("Legacy API disabled", "ApiRequestHandler");
+#endif
 }
 
 httpd::Response ApiRequestHandler::onRequest(const httpd::Request& request)
--- a/bridge/ApiRequestHandler.cpp
+++ b/bridge/ApiRequestHandler.cpp
@@ -381,11 +381,6 @@ ApiRequestHandler::ApiRequestHandler(bridge::MixerManager& mixerManager, transpo
       _legacyApiRequestHandler(std::make_unique<LegacyApiRequestHandler>(mixerManager, sslDtls))
 #endif
 {
-#if ENABLE_LEGACY_API
-    logger::logAlways("Legacy colibri API enabled", "ApiRequestHandler");
-#else
-    logger::logAlways("Legacy API disabled", "ApiRequestHandler");
-#endif
 }
 
 httpd::Response ApiRequestHandler::onRequest(const httpd::Request& request)
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -3198,13 +3198,13 @@ public:
     SetMaxMediaBitrateJob(transport::RtcTransport& transport,
         uint32_t reporterSsrc,
         uint32_t ssrc,
-        uint32_t bitrate,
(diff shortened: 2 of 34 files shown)


bridge/ApiRequestHandler.cpp
```diff
@@ -381,6 +381,11 @@ ApiRequestHandler::ApiRequestHandler(bridge::MixerManager& mixerManager, transpo
       _legacyApiRequestHandler(std::make_unique<LegacyApiRequestHandler>(mixerManager, sslDtls))
 #endif
 {
+#if ENABLE_LEGACY_API
+    logger::logAlways("Legacy colibri API enabled", "ApiRequestHandler");
+#else
+    logger::logAlways("Legacy API disabled", "ApiRequestHandler");
+#endif
```
marcusspangenberg, line 388: Could you remove this? I understand that you needed it at first, but legacy api is for Symphony internal builds only.

bridge/engine/EngineMixer.cpp
```diff
@@ -3190,4 +3192,100 @@ void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
     }
 }
 
+class SetMaxMediaBitrateJob : public jobmanager::CountedJob
```
marcusspangenberg, line 3195: Please move this to the anonymous namespace in the cpp file, or a separate file.

bridge/engine/EngineMixer.cpp
```diff
@@ -3190,4 +3192,100 @@ void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
...
+    transport::RtcTransport& _transport;
+    const uint32_t _ssrc;
+    const uint32_t _reporterSsrc;
+    const uint32_t _bitrate;
+    memory::PacketPoolAllocator& _allocator;
+};
+
+void EngineMixer::checkVideoBandwidth(const uint64_t timestamp)
+{
+    if (utils::Time::diffGE(_lastVideoBandwidthCheck, timestamp, utils::Time::sec * 3))
```
marcusspangenberg, line 3236: Could you invert this with return instead, to avoid having everything in the function in an extra scope?

bridge/engine/EngineMixer.cpp
```diff
@@ -3190,4 +3192,100 @@ void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
...
+            }
+        }
+
+        minUplinkEstimate = std::max(minUplinkEstimate, _config.slides.minBitrate.get());
+
+        if (presenterSimulcastLevel)
+        {
+            const uint32_t slidesLimit = minUplinkEstimate * _config.slides.allocFactor;
+
+            logger::debug("limiting bitrate for ssrc %u, at %u",
```
marcusspangenberg, line 3268: Could this be useful as an info level log or will it be too frequent?

marcusspangenberg:
Do we need to start negotiating max media bit rate feedback in the SDP after this and possibly handle an answer with it disabled?

bwe/RateController.cpp
```diff
@@ -13,6 +15,33 @@ namespace bwe
 {
 const uint32_t ntp32Second = 0x10000u;
 
+uint32_t calculateTargetQueue(double bandwidthKbps, uint32_t minRttNtp, const RateControllerConfig& config)
```
marcusspangenberg, line 18: If this is local to this cpp file, could you move it to the anonymous namespace instead?

PR #37, opened 2021-12-16
OlofkRctlLog

olofkallander:
* RC probes spread in time to avoid all packets being sent before 2nd SR.
* Probe applied on current target queue in case target queue is already consumed by pending data from streams.

### DIFF
--- a/bwe/RateController.cpp
+++ b/bwe/RateController.cpp
@@ -648,7 +648,7 @@ size_t RateController::getPacingBudget(uint64_t timestamp) const
     {
         currentTargetQ += _config.mtu;
         currentTargetQ +=
-            static_cast<uint32_t>(_model.targetQueue * 2 * (timestamp - _probe.start) / _config.probeDuration);
+            static_cast<uint32_t>(_model.targetQueue * 3 * (timestamp - _probe.start) / _config.probeDuration);
     }
 
     return currentTargetQ - std::min(currentTargetQ, currentQueue);
--- a/bwe/RateController.cpp
+++ b/bwe/RateController.cpp
@@ -30,32 +30,23 @@ uint32_t calculateTargetQueue(double bandwidthKbps, uint32_t minRttNtp, const bw
 namespace bwe
 {
 
-double RateController::BacklogAnalysis::getBitrateKbps() const
+uint32_t RateController::BacklogAnalysis::getBitrateKbps() const
 {
     if (delaySinceSR == 0)
     {
         return 0;
     }
 
-    return static_cast<double>(receivedAfterSR) * ntp32Second / (125.0 * delaySinceSR);
+    return utils::checkedCast<uint32_t>(static_cast<double>(receivedAfterSR) * ntp32Second / (125.0 * delaySinceSR));
 }
 
-double RateController::BacklogAnalysis::getSendRateKbps() const
+uint32_t RateController::BacklogAnalysis::getSendRateKbps() const
 {
     if (transmitPeriod == 0)
     {
         return 0;
     }
-    return static_cast<double>(receivedAfterSR) * utils::Time::ms * 8 / transmitPeriod;
-}
-
-inline void RateController::Model::onPacketSent(uint64_t timestamp, uint16_t size)
-{
-    queue -= std::min(queue,
-        static_cast<uint32_t>(utils::Time::diff(lastTransmission, timestamp) * bandwidthKbps / (8 * utils::Time::ms)));
-
-    queue += size;
-    lastTransmission = timestamp;
+    return utils::checkedCast<uint32_t>(static_cast<double>(receivedAfterSR) * utils::Time::ms * 8 / transmitPeriod);
 }
 
 RateController::RateController(size_t instanceId, const RateControllerConfig& config)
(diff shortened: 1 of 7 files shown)


bwe/RateController.cpp
```diff
@@ -298,12 +327,24 @@ RateController::BacklogAnalysis RateController::analyzeBacklog(uint32_t reportNt
             item.delaySinceSR > ntp32Second / 20)
         {
             auto reportCandidate = analyzeProbe(i, modelBandwidthKbps);
-            if (isGood(reportCandidate) || report.lossCount > 0)
+            if (reportCandidate.lossCount > 0 && report.empty())
```
RicardoMDomingues, line 330: Why we stop as soon we have a candidate loss count? If we already have a goodReport when detect another with a lost count (assuming a good report is not not empty), we continue to search. But if we have detect a loss on the 1st candidate, we will abort the search

olofkallander: The reasoning is that if we find a loss report it is the most recent condition. In history there may be another report with no loss and if we pick the better one, we will not back off if we detect loss. At least not until the good report has been shifted out.

bwe/RateController.cpp
```diff
@@ -611,7 +656,24 @@ size_t RateController::getPacingBudget(uint64_t timestamp) const
             static_cast<uint32_t>(
                 _model.bandwidthKbps * (timestamp - _model.lastTransmission) / (8 * utils::Time::ms)));
 
-    return _model.targetQueue - std::min(_model.targetQueue, currentQueue);
+    auto currentTargetQ = _model.targetQueue;
+    if (_probe.isProbing(timestamp))
+    {
+        currentTargetQ += _config.mtu;
+        currentTargetQ += static_cast<uint32_t>(_model.targetQueue * (timestamp - _probe.start) / _config.probeRampup);
```
RicardoMDomingues, line 663: We said when we probe we want to double the targetQ. But we can go much further than that.
A probe can last 700ms. if _config.probeRampup is 300ms, the we have the double queue at 300ms . And we will have 3x more at 600ms. Shouldn't we limit it to the double? or isn't a problem having such high budget?

olofkallander: True we are not limiting and that is perhaps why the bw probes worked better now.
The risk is that we exceed the buffer capacity of the network and induce loss. The is not a risk when we are under estimating since the drain rate is much higer than we think. But it is a problem when we are already near capacity. Then the probe will cause build up and then loss, and it will take much longer time to reduce the delay. It can only drain as fast as we are underutilizing the link. If we have 98% utilization the delay will remain for a very long time.

bwe/RateController.cpp
RicardoMDomingues: Just to confirm if I understood.

Receive 2 reports:
1. with loss
2. no loss 

1. is the last condition and we should react even if it is not best report according to `bestReport` method).

Receive 2 reports:
1. no loss -> (better report according to `bestReport` method
2. loss

We will not consider the loss report but is not to important because is not the most recent condition (meaning that that lost could be only a small glitch since the 2nd which is a better report and the most recent condition has not reported a lost).

Did I understood the logic correctly?

olofkallander: yes, it is problem right if we receive loss, react to that. Then there is another good report. Now that we search for better reports the loss will be picked again. We will not act on it since there is a check later if the report is the same as the one we acted on previously.
Before we just checked the last report, but that one is likely to have a lot of unconfirmed packets. A previous report may on the other hand have a lot more confirmed packets.

Perhaps we need to reconsider the lagorithm.

bwe/RateController.cpp
```diff
@@ -298,12 +327,24 @@ RateController::BacklogAnalysis RateController::analyzeBacklog(uint32_t reportNt
             item.delaySinceSR > ntp32Second / 20)
         {
             auto reportCandidate = analyzeProbe(i, modelBandwidthKbps);
```
RicardoMDomingues, line 329:
```suggestion
     auto reportCandidate = analyzeProbe(i, modelBandwidthKbps);
     if (reportCandidate.senderReportItem->reportNtp == _lastProbe.ntp &&
        reportCandidate.delaySinceSR == _lastProbe.delaySinceSR &&
        reportCandidate.packetsReceived == _lastProbe.packetsReceived)
    {
       // From now on we are only wasting CP and probably ignoring a good recent by choosing one that already was
       // processed
        return report;
    }

```

RicardoMDomingues: I think here we are potential ignore newer candidates if we find a older one that will be ignored because it was already utilised. Maybe we should mark a report as analysed and ignore it

Imagine in backlog I have:
[report2]
[report1]


and we decide that report1 is better than 2. Then we use the 1.

Next I have:
[report3]
[report2]
[report1]
 
And I decide to choose again the 1. But the report3 is more recent and we already have used the 1. Shouldn't we use the 3?

RicardoMDomingues: For me also make sense mark the 2 as analysed. Because if we prefer the report 1 over 2, doesn't make sense in next iteration use the report 2 just because the 1 is not in the backlog any longer

olofkallander: I agree, I was also thinking about marking which reports we already used. A problem with that is that the report A may be the best atm, then another B has more confirmed packets, then another RB confirms more packets and the A is actually the best again. But we cannot use A because it was marked as used.

olofkallander: I am tempted to add the improvement you suggest above. The rationale would be that if a later RB comes in, it should confirm packets also in that last best reports and other reports. We would then recalculate all reports and pick the best.

But what if _lastProbe does not contain any packets from the RB ssrc? What if it is an RB for probe packets that are important but sits right after a previous SR?
We would not look beyond the _lastProbe then. Maybe it is unusual but we need to resolve that.
It could be done when analysing the RB. That we clear a mark or something in lastProbe if the RB affects any packets before lastProbe.

PR #38, opened 2022-01-11
Ricardo/rctl

RicardoMDomingues:


bwe/RateController.cpp
```diff
@@ -100,7 +100,7 @@ void RateController::onSenderReportSent(uint64_t timestamp, uint32_t ssrc, uint3
 
     const auto timeToProbe = _probe.start == 0 || utils::Time::diffGE(_probe.start, timestamp, _probe.interval);
 
-    if (_model.queue.size() < _model.targetQueue && ((_probe.count < 5 && _probe.duration == 0) || timeToProbe))
+    if (_model.queue.size() < _model.targetQueue && ((_probe.count < 5 && _probe.duration == 0) || (timeToProbe && !hasBackedOffDueToLostRecently(timestamp))))
```
RicardoMDomingues, line 103: Do not start a probe if we have backed off the estimation due a lost recently

bwe/RateController.cpp
```diff
@@ -167,9 +167,7 @@ void RateController::onReportBlockReceived(uint32_t ssrc,
     {
         if (!_lastProbe.empty() && item.reportNtp == _lastProbe.ntp && item.type == PacketMetaData::SR)
         {
-            // probes before last probe may have been updated.
-            _lastProbe.packetsReceived = 0;
-            _lastProbe.ntp = 0;
+            _mayHaveBetterProbeThanLast = true;
```
RicardoMDomingues, line 170: Do not throw away the _lastProbe. Just mark as we may have better probes than the last to be consider

bwe/RateController.cpp
```diff
@@ -226,15 +224,22 @@ bool isProbe(const T& backlog, double bandwidthKbps, uint32_t trainEnd, uint32_t
     for (uint32_t i = srIndex - 1;; --i)
     {
         const auto& item = backlog[i];
-        queue.onPacketSent(item.transmissionTime, item.size);
-        if (queue.size() <= item.size)
+
+        // It's important to drain and check before add packet. Otherwise we can't be sure we went
+        // lower than initial size because adding a new packet in the most of cases will increase the queue
+        // above initial
+        queue.drain(item.transmissionTime);
```
RicardoMDomingues, line 231: drain first! otherwise we just need the packets added to have a bigger size that the SR

bwe/RateController.cpp
olofkallander: Are we right about this lastProbe, regardless of this change?
If we have SR1 .... SR2....RB2a...RB1...RB2b, we may have selected SR2 as last probe but since we confirmed packets  between RB2a and RB2b, we may have improved both SR1 and SR2.

RicardoMDomingues: Yes, it why mark this as "_mayHaveBetterProbeThanLast" and this will required to analyse all backlog (not only until _lastProbe) . Then we are giving the opportunity to SR2 to be analysed again. And even if it SR1 os chosen again, it only be considered the same if it has confirmed the same number of the last time.

olofkallander: I was just thinking it will cause a deeper search if we get a report referencing to the previously selected probe, but if the report is affecting both the last and previous probe we do not examine both. At RB1 for example.

bwe/RateController.cpp
```diff
@@ -167,9 +167,7 @@ void RateController::onReportBlockReceived(uint32_t ssrc,
     {
         if (!_lastProbe.empty() && item.reportNtp == _lastProbe.ntp && item.type == PacketMetaData::SR)
```
olofkallander, line 168: What if the RB referes to another previous SR that is now better confirmed along with the _lastProbe. 
We only search deeper if the RB refers to this lastProbe.

bwe/RateController.cpp
```diff
@@ -112,9 +167,14 @@ void RateController::onSenderReportSent(uint64_t timestamp, uint32_t ssrc, uint3
...
             reportNtp);
-        if (_probe.count == 10 || _probe.count == 30)
+
+        const auto probesSinceLastReduction = _probe.count - _probe.countOnLastIntervalReduction;
+
+        if (_probe.interval < Probe::MAX_INTERVAL && (probesSinceLastReduction == 10 || probesSinceLastReduction == 30))
         {
-            _probe.interval *= 2;
+            // New variable to shut up compiler warnings on mac
+            constexpr auto maxInterval = Probe::MAX_INTERVAL;
```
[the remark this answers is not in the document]
RicardoMDomingues: Actually the comment is wrong. it was not a compiler warning but linker error. It says it can find the MAX_INTERVAL when I give it as parameter of `std::min`. `MAX_INTERVAL` is a `constexpr` and the value is assign in declaration, I am being used this a lot in C++17 and it works well but in C++14 a `static constexpr` member is not inlined (at least not in all cases). When I assign directly or use id a `if` condition, the compiler will use the literal value directly instead of reference something to be solved by the linker. But when I give it as an argument the compiler it will reference some entry to be linked.
But I agree the type should be explicitly to the be easiest to read the code since it is not explicit by the rhs of the assign.

bwe/RateController.cpp
```diff
@@ -555,9 +680,9 @@ void RateController::onReportReceived(uint64_t timestamp,
             _model.queue.setBandwidth(currentEstimate - 0.1 * (currentEstimate - receiveRateKbps));
         }
     }
-    else if (!sameProbe && backlogReport.lossCount > 2 && backlogReport.lossRatio > 0.02 &&
+    else if (backlogReport.lossCount > 2 && backlogReport.lossRatio > 0.02 &&
```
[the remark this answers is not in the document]
RicardoMDomingues: I think it is. As you said the lossRation should be low that `2%`. Then this should be just to ensure that a not good report with a very short period will not reduce the estimation just because it lost 2 packets and in that short period it represents more than 2% loss. Here we are ensure we only consider probes bigger that 150ms (where 3 losses represent 2%) or shorter probes if the lossCount is something significant (3 or more sound reasonable). For the long probes (more than 150ms) the condition is always more than 2% that also sounds very reasonable

bwe/RateController.cpp
```diff
@@ -18,6 +18,25 @@
 namespace
 {
 const uint32_t ntp32Second = 0x10000u;
+const uint64_t logProbeIssuesInterval = utils::Time::sec * 30;
+const uint64_t logEstimationModelInterval = utils::Time::sec * 30;
+
+enum ProbeIssues {
+    NO_ISSUES = 0x0,
```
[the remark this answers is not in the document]
RicardoMDomingues: SUCCESS OR EXCELLENT also not sound correct for `ProbeIssues`. Maybe the name of enum is the problem. Maybe should be something like `ProveEvaluation`: EXCELLENT, NO_REPORT_FOUND, etc. wdyd?

bwe/RateController.cpp
```diff
@@ -576,6 +671,30 @@ void RateController::onReportReceived(uint64_t timestamp,
         _model.queue.setSize(backlogReport.networkQueue);
     }
 
+    const bool isTimeToLogModel = _modelLogInfo._lastLoggedTimestamp == 0 || utils::Time::diffGE(_modelLogInfo._lastLoggedTimestamp, timestamp, logEstimationModelInterval);
```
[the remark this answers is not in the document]
RicardoMDomingues: That is not the last version. I decided it to remove that in one of the previous commits as I didn't find it that useful. Do you want me to put it again?

bwe/RateController.cpp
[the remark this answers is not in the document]
RicardoMDomingues: I changed the name from `ProbeIssues` to `ProveEvaluation`. i've changed also the method name from `getProbeIssues` to `evaluateProbe` and create an anonymous function to update `_probeMetrics` to try keep the code in `analyzeBacklog` as clean as possible

bwe/RateController.cpp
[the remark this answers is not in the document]
RicardoMDomingues: I think it is. > 2 is just to ensure we don't react because 1 or 2 loss in short probes. where it can represent more than 2%. In the long probes is expected to have a lot of packets. Then probably 3 packets less than 2% in the most of the cases.

olofkallander (approved):

PR #39, opened 2022-01-19
RTC-12054 Add logging for srtp unprotect errors

pcarlsym:


bridge/engine/VideoForwarderReceiveJob.cpp
```diff
@@ -87,10 +87,16 @@ void VideoForwarderReceiveJob::run()
 
     if (!_sender->unprotect(_packet))
     {
-        logger::error("Failed to unprotect srtp %s, ssrc %u, mixer %s",
+        const auto header = rtp::RtpHeader::fromPacket(*_packet);
```
olofkallander, line 90: It really should be an rtp packet, but can we check that header is not null before logging?

RicardoMDomingues (approved):

PR #41, opened 2022-01-20
Reduce mem usage in packet pools

olofkallander:
Reduced packet size in packet pools cuts memory usage in half. the packet caches are more numerous and saves most.
Audio specific pool is declared to handle PCM16 packets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,6 +234,7 @@ set(FILES
         memory/List.h
         memory/Packet.h
         memory/PacketPoolAllocator.h
+        memory/AudioPacketPoolAllocator.h
         memory/PoolAllocator.h
         memory/PriorityQueue.h
         memory/RefCountedPacket.h
--- a/bridge/engine/AudioForwarderReceiveJob.cpp
+++ b/bridge/engine/AudioForwarderReceiveJob.cpp
@@ -17,7 +17,7 @@ void AudioForwarderReceiveJob::onPacketDecoded(const int32_t decodedFrames, cons
 {
     if (decodedFrames > 0)
     {
-        auto pcmPacket = memory::makePacket(_allocator, *_packet);
+        auto pcmPacket = memory::makePacket(_audioPacketAllocator, *_packet);
         if (!pcmPacket)
         {
             return;
@@ -27,7 +27,7 @@ void AudioForwarderReceiveJob::onPacketDecoded(const int32_t decodedFrames, cons
         memcpy(rtpHeader->getPayload(), decodedData, decodedPayloadLength);
         pcmPacket->setLength(rtpHeader->headerLength() + decodedPayloadLength);
 
-        _engineMixer.onMixerAudioRtpPacketDecoded(_sender, pcmPacket, _allocator);
+        _engineMixer.onMixerAudioRtpPacketDecoded(_sender, pcmPacket, _audioPacketAllocator);
         return;
     }
 
@@ -55,7 +55,7 @@ void AudioForwarderReceiveJob::decodeOpus(const memory::Packet& opusPacket)
         return;
     }
 
-    uint8_t decodedData[memory::Packet::size];
+    uint8_t decodedData[memory::AudioPacket::size];
     auto rtpPacket = rtp::RtpHeader::fromPacket(*_packet);
     if (!rtpPacket)
     {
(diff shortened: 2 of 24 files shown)


olofkallander:
I will update this PR with e2e test for mixer mode to make sure the AudioPackets are used correctly.

bridge/engine/EncodeJob.cpp
```diff
@@ -75,32 +77,31 @@ void EncodeJob::run()
...
-        _packet->setLength(headerLength + encodedBytes);
-        rtpPacket->ssrc = _outboundContext._ssrc;
-        rtpPacket->timestamp = (_rtpTimestamp * 48llu) & 0xFFFFFFFFllu;
-        rtpPacket->sequenceNumber = _outboundContext._sequenceCounter++ & 0xFFFFu;
-        rtpPacket->payloadType = targetFormat._payloadType;
-        doSend(_packet);
+        {
+            auto opusPacket = memory::makePacket(_outboundContext._allocator, rtpPacket, rtpPacket->headerLength());
+            auto opusHeader = rtp::RtpHeader::fromPacket(*opusPacket);
+            std::memcpy(opusHeader->getPayload(), encodedData, encodedBytes);
```
RicardoMDomingues, line 88: I am wondering if we can't opus packet (`makePacket`) before call `opusEncoder->encode` and give opusHeader->getPayload() as encodedDataOutput, in the end is less one mem copy

RicardoMDomingues (approved):

bridge/engine/EncodeJob.cpp
olofkallander: yes, we should be able to do that

RicardoMDomingues (approved):

CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ elseif(UNIX)
     set(CMAKE_INSTALL_RPATH "./libs")
     set(CMAKE_BUILD_WITH_INSTALL_RPATH ON)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -lc++abi -lrt")
+    link_directories("/usr/local/lib")
```
RicardoMDomingues, line 48: Not sure if we should commit absolute paths. Please take a look on PR #43. Even now if we have this open source, we shouldn't force paths for our specific envs. On other hand, this is a very common path on linux, shouldn't hurt if it stay

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -66,50 +66,40 @@ void AudioForwarderReceiveJob::decodeOpus(const memory::Packet& opusPacket)
...
                 "OpusDecodeJob",
-                sequenceNumber,
-                expectedSequenceNumber);
+                _extendedSequenceNumber,
+                decoder.getExpectedSequenceNumber());
             return;
         }
-
-        if (sequenceNumber > expectedSequenceNumber)
+        else
```
RicardoMDomingues, line 80: In my opinion if we already have a early return on the 1st `if` we should not create an `else` block. This increase the indentation and make it harder to read

RicardoMDomingues (approved):
LGTM. Just some suggestion to improve code readability

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -66,55 +66,46 @@ void AudioForwarderReceiveJob::decodeOpus(const memory::Packet& opusPacket)
...
             logger::debug("Lost opus packet sequence %u expected %u, fec",
                 "OpusDecodeJob",
-                sequenceNumber,
-                expectedSequenceNumber);
-            const auto lastPacketDuration = decoder.getLastPacketDuration();
+                _extendedSequenceNumber,
+                decoder.getExpectedSequenceNumber());
 
-            for (auto i = expectedSequenceNumber; i < sequenceNumber - 2; ++i)
+            for (uint32_t i = 0; i < _extendedSequenceNumber - decoder.getExpectedSequenceNumber() - 1 && i < 5; ++i)
```
RicardoMDomingues, line 87:
```suggestion
const auto concealTimes = std::min(5, _extendedSequenceNumber - decoder.getExpectedSequenceNumber() - 1);
for (uint32_t i = 0; i < concealTimes; ++i)
```

I think it make it more readable

bridge/engine/AudioForwarderReceiveJob.cpp
olofkallander: of course

CMakeLists.txt
olofkallander: I plan to replace with your other PR. I was not sure to commit this since it compiled before on Jenkins, but not on my machine. I can keep it locally until replaced.

olofkallander (commented):
It seems the test does not work on all platforms. Also, the number of amplitude records vary so I have to work on the test verification conditions

RicardoMDomingues (approved):

PR #42, opened 2022-01-24
Choose a recording endpoint with same ip family than FWD

RicardoMDomingues:
smb can open an ipv4 and an ipv6 interface for recording and we were just round robin to choose one interface without care about FWD ip family. This cause empty records since smb couldn't send packets to FWD when the ip family mismatch

### DIFF
--- a/transport/sctp/SctpAssociationImpl.cpp
+++ b/transport/sctp/SctpAssociationImpl.cpp
@@ -8,7 +8,16 @@
 #include "utils/Time.h"
 #include <set>
 
-#define SCTP_LOG(fmt, ...) // logger::debug(fmt, ##__VA_ARGS__)
+#define SCTP_LOG_ENABLE 0
+
+#if SCTP_LOG_ENABLE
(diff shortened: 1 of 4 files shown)


transport/sctp/SctpAssociationImpl.cpp
```diff
@@ -8,7 +8,16 @@
...
 #include <set>
 
-#define SCTP_LOG(fmt, ...) // logger::debug(fmt, ##__VA_ARGS__)
+#define SCTP_LOG_ENABLE 0
+
+#if SCTP_LOG_ENABLE
+    #define SCTP_LOG(fmt, ...) logger::debug(fmt, ##__VA_ARGS__)
+#else
+    // silence compiler warnings
+    #define SCTP_LOG(fmt, logId, ...) (void)logId;
```
RicardoMDomingues, line 17: Not related changes, only to silence compiler warnings about unused variables

pcarlsym (approved):

PR #46, opened 2022-02-23
Avoid access to nullptr when packet allocators run out of memory

RicardoMDomingues:
We got a crash on SMB on load tests where we can see in the logs that some allocators as has run out of memory. This can provoke SEGV as we are not checking if the allocated memory pointer is a nullptr.
Probably we will have a lot of issues on this conferences where this happen but at least smb shouldn't crash


transport/recp/RecStreamAddedEventBuilder.cpp
```diff
@@ -47,22 +47,26 @@ RecStreamAddedEventBuilder& RecStreamAddedEventBuilder::setBridgeCodecNumber(uin
...
-    // Move the fields after the endpoint size to the right position
-    std::memmove(packet->get() + newEndPosPlusPadding, packet->get() + oldEndPosPlusPadding, bytesAfterEndpoint);
-    // clean new padding (We will add always a padding to 32 to 32bits even if we don't have bytes after endpoint)
-    std::memset(packet->get() + newEndPos, 0, newEndPosPlusPadding - newEndPos);
-    packet->setLength(newEndPosPlusPadding + bytesAfterEndpoint);
-
-    castPacket(packet)->endpointLen = static_cast<uint16_t>(endpoint.size());
-    std::memcpy(getEndpointBuffRef(packet), endpoint.c_str(), endpoint.size());
+    auto* packet = getPacket();
+    if (packet)
```
olofkallander, line 51: should we do 
if (!packet){return *this;}

transport/recp/RecStreamAddedEventBuilder.cpp
```diff
@@ -72,13 +76,17 @@ RecStreamAddedEventBuilder& RecStreamAddedEventBuilder::setWallClock(std::chrono
...
-    const uint32_t endpointEndPos = RecStreamAddedEventBuilder::MinSize + castPacket(getPacket())->endpointLen;
-    // Wall clock position must be 32 bits aligned (we don't need to clean the padding, padding should be inserted and
-    // cleaned by setEndpoint() method)
-    const uint32_t wallClockPosition = endpointEndPos + 3u & ~3u;
-    const nwuint64_t ntpTimestamp(utils::Time::toNtp(wallClock));
-    auto packet = getPacket();
-    packet->setLength(wallClockPosition + sizeof(nwuint64_t));
-    std::memcpy(packet->get() + wallClockPosition, &ntpTimestamp, sizeof(ntpTimestamp));
+    auto* packet = getPacket();
+    if (packet)
```
olofkallander, line 80: perhaps return on !packet too

olofkallander (changes requested):
My only objection is that in some methods when we return right after the if block, the indentation is less if make a check on if(!packet) {log; return;}

olofkallander (approved):

pcarlsym (approved):

transport/BaseUdpEndpoint.h
```diff
@@ -46,6 +46,8 @@ class BaseUdpEndpoint : public Endpoint, RtcePoll::IEventListener
 
     ice::TransportType getTransportType() const override { return ice::TransportType::UDP; }
 
+    EndpointMetrics getMetrics() const final { return EndpointMetrics(_receiveJobs.getCount(), _sendJobs.getCount()); }
```
olofkallander, line 49: The send and receive jobs should be about one or two in queue.
In addition, they will show up in the common job queue count that we already have a metrics for.

What we need is number o packets waiting in _sendQueue

transport/EndpointMetrics.h
```diff
@@ -0,0 +1,30 @@
...
+    EndpointMetrics() : EndpointMetrics(0, 0) { }
+
+    EndpointMetrics& operator+=(const EndpointMetrics& rhs)
+    {
+        rxQueue += rhs.rxQueue;
+        txQueue += rhs.txQueue;
+        return *this;
+    }
+
+    uint32_t rxQueue;
```
olofkallander, line 21: we should add sendQueue here

RicardoMDomingues: I added, it's the txQueue. I can change the name if you prefer

RicardoMDomingues: Ok. I get it now

transport/EndpointMetrics.h
```diff
@@ -0,0 +1,33 @@
...
+    EndpointMetrics& operator+=(const EndpointMetrics& rhs)
+    {
+        sendQueue += rhs.sendQueue;
+        receiveKbps += rhs.receiveKbps;
+        sendKbps += rhs.sendKbps;
+        return *this;
+    }
+
+    uint32_t sendQueue;
+    double receiveKbps;
```
olofkallander, line 24: I wonder if it is better to store in kbps. No need for the precision I think.

bridge/Stats.h
```diff
@@ -52,6 +52,9 @@ struct MixerManagerStats
 
     uint32_t _receivePoolSize = 0;
     uint32_t _sendPoolSize = 0;
+    uint32_t _udpSharedEndpointsSendQueue = 0;
+    double _udpSharedEndpointsReceiveKbps = 0.0;
```
olofkallander, line 56: This could be uint32_t too

olofkallander (approved):

transport/EndpointMetrics.h
RicardoMDomingues: Maybe you were seeing my old branch where I store in bits/s? It's stored in kbps since I add this change to this branch. Or am I getting something wrong?

PR #47, opened 2022-02-24
Remove JobQueue from EngineRecordingStream and use transport jobQueue instead

RicardoMDomingues:
**Details**
Sometimes the E2E test to Cloud Recording  fails. This is because sometimes SMB does not send the StopRecordingEvent. FWD waits for this event to end the recording until report to CS the recordings that are running and CS force FWD to end the recording because the conference is no longer available. This will create a bigger recording than what is expected.

SMB didn't send the StopRecordingEvent sometimes because `RecordingEngineStream` has it own `JobQueue` to run jobs before handover the packet to `RecordingTransport` and enqueue other job on `RecordingTransport`'s `JobQueue`.
This cause sometimes the `RecordingTransport` to shutdown before the `StopRecordingEvent` has been enqueue in `RecordingTransport`'s `JobQueue`, then that packet is not sent.
To fix the problem we have removed the `JobQueue` of `RecordingEngineStream` as it only increase complexity in the moment of removing `RecordingEngineStream` and create extra context switching. Now we are using directly `JobQueue` of `RecordingTransport` when we need to create a Job in EngineMixer and that will ensure all jobs in the queue are executed before shutdown `RecordingTransport`

### DIFF
--- a/bridge/engine/EngineRecordingStream.h
+++ b/bridge/engine/EngineRecordingStream.h
@@ -16,8 +16,6 @@ struct EngineRecordingStream
         bool isAudioEnabled,
         bool isVideoEnabled,
         bool isScreenSharingEnabled,
-        jobmanager::JobManager& jobManager,
-        std::atomic_uint32_t& jobsCounter,
         PacketCache& recordingEventPacketCache)
         : _id(id),
           _endpointIdHash(endpointIdHash),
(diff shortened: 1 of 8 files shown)


bridge/engine/EngineRecordingStream.h
```diff
@@ -16,8 +16,6 @@ struct EngineRecordingStream
         bool isAudioEnabled,
         bool isVideoEnabled,
         bool isScreenSharingEnabled,
-        jobmanager::JobManager& jobManager,
```
olofkallander, line 19: super

olofkallander (approved):

PR #52, opened 2022-03-10
Improve how we stop transport to ensure we don't lose recording events

RicardoMDomingues:
Sometimes we miss the final recording event. This may happened because `waitForPendingJobs` was stopping the transport. Regular transports are stopped before this method is called and for recordings we don't want to stop them too soon. Now the `waitForPendingJobs` does not stop transports

Also made some code improvements and add logs

### DIFF
--- a/transport/RecordingTransport.cpp
+++ b/transport/RecordingTransport.cpp
@@ -113,7 +113,7 @@ void RecordingTransport::protectAndSend(memory::Packet* packet, memory::PacketPo
         // Recording events are very important to convertor and must not be lost!
         if (recp::isRecPacket(*packet))
         {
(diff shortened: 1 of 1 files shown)


pcarlsym (approved):

bridge/Mixer.cpp
```diff
@@ -101,9 +101,9 @@ bool waitForPendingJobs(const uint32_t timeoutMs, const uint32_t pollIntervalMs,
     while (totalSleepTimeMs < timeoutMs && transport.hasPendingJobs())
     {
         totalSleepTimeMs += pollIntervalMs;
-        transport.stop();
```
olofkallander, line 104: Good change. I am not sure it is safe to stop multiple times or a small change could cause it to be problematic.

olofkallander (approved):

PR #58, opened 2022-03-29
Improve RateController logs

RicardoMDomingues:


transport/TransportImpl.cpp
```diff
@@ -1757,16 +1757,17 @@ void TransportImpl::appendRemb(memory::Packet* rtcpPacket,
...
         logger::info("Estimates 5s, Downlink %u - %ukbps, rate %.1fkbps, Uplink rctl %.0fkbps, rate %.1fkbps, remb "
-                     "%ukbps, rtt %.1fms, pacingQ %zu",
+                     "%ukbps, rtt %.1fms, pacingQ %zu , rtpProbingEnabled %s",
             _loggableId.c_str(),
             oldMin,
             oldMax,
             _bwe->getReceiveRate(timestamp),
             _rateController.getTargetRate(),
-            _sendRateTracker.get(utils::Time::ms * 600) * 8 * utils::Time::ms,
+            _sendRateTracker.get(timestamp, utils::Time::ms * 600) * 8 * utils::Time::ms,
```
RicardoMDomingues, line 1766: Add the timestamp here because if we not use the timestamp, it will use the last one used to update the rateTracker. Then it will never back to zero again if we stop to send data

olofkallander (approved):

PR #64, opened 2022-03-31
Olofk/sendtime

olofkallander:
RTC-12618
absSendTime was not set when not first hdr ext in packet.
Parsing of hdr ext could fail.
Invalid ext ids were allowed

### DIFF
--- a/rtp/RtpHeader.cpp
+++ b/rtp/RtpHeader.cpp
@@ -50,15 +50,6 @@ RtpHeader* RtpHeader::fromPtr(void* p, const size_t len)
     return header;
 }
 
-RtpHeader* RtpHeader::create(void* p, const size_t len)
-{
-    assert(len >= MIN_RTP_HEADER_SIZE);
-    auto header = reinterpret_cast<RtpHeader*>(p);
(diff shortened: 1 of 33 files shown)


rtp/RtpHeader.cpp
```diff
@@ -101,45 +92,96 @@ void RtpHeader::setExtensions(const RtpHeaderExtension& extensions)
...
+    return utils::TlvCollection<GeneralExtension1Byteheader>(data,
+        const_cast<uint8_t*>(findExtensionsEnd(data, data + length * sizeof(uint32_t))));
 }
 
-void GeneralExtension1Byteheader::setDataLength(int length)
+bool RtpHeaderExtension::isValid() const
 {
-    if (length < 2)
+    auto extEnd = findExtensionsEnd(data, data + length * sizeof(uint32_t));
+    if (extEnd == data + length * sizeof(uint32_t))
```
RicardoMDomingues, line 156: Code simplification:
``` return (extEnd == data + length * sizeof(uint32_t)) || ((*extEnd & 0xF0) == 0xF0);

RicardoMDomingues (approved):

PR #66, opened 2022-04-01
fix ice map full

olofkallander:
RTC-12666 TCP Fallback during load
Fix for this. Main problem was that pending stun transaction map in UdpEndpoint became full. It is cleared at call shutdown but all failed candidates will create a number of transactions that are not cleared until end of call.
The fix is that stun requests, that will result in responses from ports that are routed anyway to the transport, do not need to be tracked by transacation id. Also, all transactions are cleaned up and earlier.

### DIFF
--- a/transport/UdpEndpoint.cpp
+++ b/transport/UdpEndpoint.cpp
@@ -25,6 +25,22 @@ private:
     UdpEndpoint& _endpoint;
     Endpoint::IEvents* _listener;
 };
+
+class UnRegisterStunListenerJob : public jobmanager::Job
+{
+public:
+    UnRegisterStunListenerJob(UdpEndpoint& endpoint, __uint128_t transactionId)
+        : _endpoint(endpoint),
+          _transactionId(transactionId)
+    {
+    }
+
+    void run() override { _endpoint.internalUnregisterStunListener(_transactionId); }
+
+private:
+    UdpEndpoint& _endpoint;
+    __uint128_t _transactionId;
+};
 } // namespace
 
 // When this endpoint is shared the number of registration jobs and packets in queue will be plenty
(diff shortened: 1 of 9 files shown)


transport/UdpEndpoint.cpp
```diff
@@ -142,10 +175,12 @@ void UdpEndpoint::dispatchReceivedPacket(const SocketAddress& srcAddress, memory
             listener = findListener(_iceResponseListeners, transactionId);
             if (listener)
             {
-                // TODO there is still risk of depletion as not all transactions are responded to.
-                // Only when channel is brought down will it clear its transaction listeners
                 _iceResponseListeners.erase(transactionId);
             }
+            else
+            {
+                listener = findListener(_dtlsListeners, srcAddress);
```
RicardoMDomingues, line 183: Why do we ned this? if we didn't get a response in time we should ignore that, right?

transport/ice/IceSession.cpp
```diff
@@ -1150,8 +1153,16 @@ void IceSession::CandidatePair::send(const uint64_t now)
...
     {
+        if (localEndpoint.endpoint)
+        {
+            auto& frontTransaction = _transactions.front();
+            if (!frontTransaction.acknowledged())
+            {
+                localEndpoint.endpoint->cancelStunTransaction(frontTransaction.id.get());
+            }
+        }
         _transactions.erase(_transactions.begin());
```
RicardoMDomingues, line 1166: As far as I understood. We don't erase elements in the middle, we always delete from `begin` if we need. Then `std::deque` seems more suitable for this case as it can handle the erase on beginning without copy all other elements (when `std::vector` has to move the others to begining).

The down size is `std::queue`on libc++ allocates space for 16 elements when needs only one. `std::vector` grows more granularly but when we have 10 transaction, `std::vector` probably already had allocated space to 16 elements anyway

RicardoMDomingues (commented):
https://github.com/finos/SymphonyMediaBridge/blob/fb846cc8e9bb306be545a1b64d035b69c170ce08/transport/TransportImpl.cpp#L1046

I am wondering if we should  call`TransportImpl::internalIceReceived`  if message fail to check the integrity. I think it will be validated further on but seems we are consume more resources in case of an attack.

transport/ice/IceSession.cpp
olofkallander: Yes, I will change to deque

transport/UdpEndpoint.cpp
olofkallander: I changed it so we do not ad listener for stun transaction ids if the remote port is already routed to our transport in dtlsListeners. This saves space in the _iceREsponseListeners and we would have registered the port anyway to get DTLS, RTP.

olofkallander:
Regarding internalIceReceived, I guess we could ignore it for invalid response. but for invalid request we are supposed to send a response. In that case I prefer if it is constructed in IceSession.

PR #68, opened 2022-04-04
Olofk/cpumetric

olofkallander:
Adds cpu metric for workers, manager, engine, rtce to stats.
Replaces engine idle

### DIFF
--- a/generate_localmake.sh
+++ b/generate_localmake.sh
@@ -6,4 +6,4 @@ if [ $# -eq 0 ]; then
 fi
 
 rm -rf CMakeCache.txt CMakeFiles googletest-* Makefile
-cmake -DCMAKE_BUILD_TYPE=$1 -G "Unix Makefiles" .
+cmake -DCMAKE_BUILD_TYPE=$1 -DENABLE_LEGACY_API=1 -G "Unix Makefiles" .
--- a/transport/UdpEndpoint.cpp
+++ b/transport/UdpEndpoint.cpp
@@ -52,9 +52,9 @@ UdpEndpoint::UdpEndpoint(jobmanager::JobManager& jobManager,
     RtcePoll& epoll,
     bool isShared)
     : BaseUdpEndpoint("UdpEndpoint", jobManager, maxSessionCount, allocator, localPort, epoll, isShared),
-      _iceListeners(maxSessionCount),
-      _dtlsListeners(maxSessionCount * 8),
-      _iceResponseListeners(maxSessionCount * 32)
+      _iceListeners(maxSessionCount * 2),
+      _dtlsListeners(maxSessionCount * 16),
+      _iceResponseListeners(maxSessionCount * 64)
 {
 }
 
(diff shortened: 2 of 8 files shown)


RicardoMDomingues (approved):

pcarlsym (approved):

olofkallander:
I added thread cpu stats for engine, workers (avg), rtc ePoll, and MixerManager. When any of those hit 1.0 it means the SMB cannot take on more traffic regardless if the cpu as a total is underutilized.
Workers are always evenly loaded so I report the average cpu for the workers.

I also increased the size of the pending ice requests in udpendpoint because I ran into full map warning during a load test that otherwise seemed to be working out well.

transport/UdpEndpoint.cpp
```diff
@@ -52,9 +52,9 @@ UdpEndpoint::UdpEndpoint(jobmanager::JobManager& jobManager,
     RtcePoll& epoll,
     bool isShared)
     : BaseUdpEndpoint("UdpEndpoint", jobManager, maxSessionCount, allocator, localPort, epoll, isShared),
-      _iceListeners(maxSessionCount),
-      _dtlsListeners(maxSessionCount * 8),
-      _iceResponseListeners(maxSessionCount * 4)
+      _iceListeners(maxSessionCount * 2),
```
olofkallander, line 55: We need more slots in the map than max expected otherrwise it will be inefficient with many hash collisions.

transport/UdpEndpoint.cpp
```diff
@@ -52,9 +52,9 @@ UdpEndpoint::UdpEndpoint(jobmanager::JobManager& jobManager,
...
+      _dtlsListeners(maxSessionCount * 16),
```
olofkallander, line 56: Each session can present multiple working udp candidates that we need to register as source ports.

transport/UdpEndpoint.cpp
```diff
@@ -52,9 +52,9 @@ UdpEndpoint::UdpEndpoint(jobmanager::JobManager& jobManager,
...
+      _iceResponseListeners(maxSessionCount * 64)
```
olofkallander, line 57: Each candidate that has not yet returned a response will cause a pending request, waiting for a response that has to be identified by transaction id. Multiple candidates with up to 8 pending requests per session. We try to remove items from this map as soon as we can rely on source port.

transport/ice/IceSession.cpp
```diff
@@ -1166,7 +1166,7 @@ void IceSession::CandidatePair::send(const uint64_t now)
         state = InProgress;
     }
     _transactions.push_back(transaction);
-    if (_transactions.size() > 10)
+    if (_transactions.size() > 8)
```
olofkallander, line 1169: Reduce pressure on ice transaction map in udp endpoint. Risk would be that "rtt" is so long that we have already shifted out a request 8 items back and the response cannot be routed. But the interval is increased when there is no response so it should be enough time for a response to match up with an older pnding request.

generate_localmake.sh
```diff
@@ -6,4 +6,4 @@ if [ $# -eq 0 ]; then
 fi
 
 rm -rf CMakeCache.txt CMakeFiles googletest-* Makefile
-cmake -DCMAKE_BUILD_TYPE=$1 -G "Unix Makefiles" .
+cmake -DCMAKE_BUILD_TYPE=$1 -DENABLE_LEGACY_API=1 -G "Unix Makefiles" .
```
pcarlsym, line 9: Was this intended or slipped through?

PR #70, opened 2022-04-06
-

olofkallander:
I found a couple of places where packets could leak.
There are some packet leaks in receive pool that am trying to find.
I am still looking for them.


RicardoMDomingues (approved):
Just a suggestion. I think this pattern of calling `free` is leak prone.
I think it would be better implement Packet unique pointers that calls the free automatically and handover ownership using move semantics

concurrency/MpmcQueue.h
```diff
@@ -77,7 +77,6 @@ class MpmcQueue
                 // the write cannot reach here until the state is set empty.
                 auto& entry = _elements[pos % _maxElements];
                 target = std::move(entry.value);
-                entry.value.~T();
```
RicardoMDomingues: Good thing 👍 . This seems a very danger thing to do on in auto initialised memory.
Maybe it would be even better change `entry.value` to an uninitialized memory like `std::aligned_storage<T>`  and call `T()` and `~T` explicitly.

RicardoMDomingues (approved):

memory/PacketPoolAllocator.h
```diff
@@ -50,4 +50,35 @@ inline Packet* makePacket(PacketPoolAllocator& allocator, const Packet& packet)
     return makePacket(allocator, packet.get(), packet.getLength());
 }
 
+// Be very careful with reset as the deleter is not changed if already set. You may try to deallocate
+// packet in the wrong pool.
+typedef std::unique_ptr<memory::Packet, PacketPoolAllocator::Deleter> PacketPtr;
```
RicardoMDomingues, line 55: I was wondering if the alias shouldn't be something with "unique" (example: UniquePacket)

memory/PacketPoolAllocator.h
```diff
@@ -50,4 +50,35 @@ inline Packet* makePacket(PacketPoolAllocator& allocator, const Packet& packet)
...
+
+inline PacketPtr makePacketPtr(PacketPoolAllocator& allocator)
```
RicardoMDomingues, line 57: Same here (like 'makeUniquePacket')

memory/PoolAllocator.h
```diff
@@ -39,8 +39,27 @@ class PoolAllocator
...
+    class Deleter
+    {
+    public:
+        Deleter() : _allocator(nullptr) {}
+        Deleter(PoolAllocator<ELEMENT_SIZE>* allocator) : _allocator(allocator) {}
+
+        template <typename T>
+        void operator()(T* r)
+        {
+            if (_allocator)
```
RicardoMDomingues, line 51: This should be only `nullptr` if the unique_ptr was default constructed. Then it only be possible to code hit here with a `nullptr` on `_allocator ` is when we use the `reset` method. Then I would suggest to make a validation in debug (like `assert`) just to ensure in debug we don't mess up

olofkallander:
>Maybe it would be even better change `entry.value` to an uninitialized memory like `std::aligned_storage<T>`  and call `T()` and `~T` explicitly. 

I thought about that too, but I am not so fond of that the object will be completely hidden in the debugger.
The bad part about using T is that it requires a default constructor on T.

RicardoMDomingues:
> > Maybe it would be even better change `entry.value` to an uninitialized memory like `std::aligned_storage<T>`  and call `T()` and `~T` explicitly.
> 
> I thought about that too, but I am not so fond of that the object will be completely hidden in the debugger. The bad part about using T is that it requires a default constructor on T.

There is another side effect. If the object holds ownership of dynamic memory or holds other resources because that resources are not released when we dequeue (only when the container itself is destructed or the entry reused) But I think it will not affect any of our use cases anyway

bridge/engine/EngineCommand.h
```diff
@@ -241,8 +240,18 @@ union CommandUnion
 
 struct Command
 {
+    Command& operator=(const Command&) = delete;
+    Command& operator=(Command&& rhs)
+    {
+        _type = rhs._type;
+        _command = rhs._command; // TODO would need move assignment in the substructs ?
+        _packet = std::exchange(rhs._packet, nullptr);
```
RicardoMDomingues, line 248:
```suggestion
        _packet = std::move(rhs._packet);
```

`std:: exchange` creates a temporary object. Usually is used on move assign operator and move contractor to exchange the values of native types. The other objects that also implements the move assign I think the `std::move` is preferable

bridge/engine/EngineCommand.h
```diff
@@ -241,8 +240,18 @@ union CommandUnion
...
+        return *this;
+    }
+
     Type _type;
     CommandUnion _command;
+    memory::UniquePacket _packet;
```
RicardoMDomingues, line 254: A bit said. But I guess we can't have it inside of substructs as the are unions

bridge/engine/EngineMixer.cpp
```diff
@@ -2480,15 +2393,13 @@ inline void EngineMixer::processAudioStreams()
...
-            !audioStream->_transport.getJobQueue().addJob<EncodeJob>(audioPacket,
-                _audioAllocator,
+            !audioStream->_transport.getJobQueue().addJob<EncodeJob>(std::move(audioPacket),
                 *ssrcContext,
                 audioStream->_transport,
                 _rtpTimestampSource,
                 audioStream->_audioLevelExtensionId,
                 audioStream->_absSendTimeExtensionId))
         {
-            _audioAllocator.free(audioPacket);
```
RicardoMDomingues, line 2491: Empty `if` block

bridge/engine/EngineMixer.h
```diff
@@ -177,50 +170,60 @@ class EngineMixer : public transport::DataReceiver
...
-        inline void lockOwner() const
+        IncomingPacketAggregate& operator=(IncomingPacketAggregate&& rhs)
         {
-            if (_transport)
-            {
-                ++_transport->getJobCounter();
-            }
+            release();
+
+            _packet = std::exchange(rhs._packet, nullptr);
```
RicardoMDomingues, line 210:
```suggestion
            _packet = std::move(rhs._packet);
```

bridge/engine/EngineMixer.h
```diff
@@ -177,50 +170,60 @@ class EngineMixer : public transport::DataReceiver
...
+            assert(_packet);
+            lockOwner();
         }
 
-        PacketT* _packet;
-        AllocatorT* _allocator;
-        transport::RtcTransport* _transport;
-        uint32_t _extendedSequenceNumber;
+        explicit IncomingPacketAggregate(IncomingPacketAggregate&& rhs)
+            : _packet(std::exchange(rhs._packet, nullptr)),
```
RicardoMDomingues, line 200:
```suggestion
            : _packet(std::move(rhs._packet)),
```

RicardoMDomingues (approved):
LGTM. just minor comments

bridge/engine/EngineCommand.h
```diff
@@ -241,8 +240,18 @@ union CommandUnion
 
 struct Command
 {
+    Command& operator=(const Command&) = delete;
+    Command& operator=(Command&& rhs)
+    {
+        _type = rhs._type;
+        _command = rhs._command; // TODO would need move assignment in the substructs ?
```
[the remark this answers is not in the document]
RicardoMDomingues: Yes, I agree. Maybe we should thinking migrate to  C++17 and replace `union` for `std::variant` or implement our version of `std::variant` but does not seems a change to be made on this PR

bridge/engine/PacketCache.cpp
```diff
@@ -68,41 +56,35 @@ bool PacketCache::add(memory::Packet* packet, const uint16_t sequenceNumber)
...
+        const auto popResult = _arrivalQueue.pop(sequenceNumberToRemove);
         assert(popResult);
 
-        auto removedPacketItr = _cache.find(removedSequenceNumberEntry);
+        auto removedPacketItr = _cache.find(sequenceNumberToRemove);
         if (removedPacketItr != _cache.end())
         {
-            removedPacketItr->second.release();
-            _cache.erase(removedSequenceNumberEntry);
+            removedPacketItr->second.reset();
```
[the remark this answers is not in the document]
RicardoMDomingues: Nice 👍

PR #72, opened 2022-04-09
Use move semantics on IncomingPacketAggregate

RicardoMDomingues:


bridge/engine/EngineMixer.h
```diff
@@ -205,22 +207,45 @@ class EngineMixer : public transport::DataReceiver
...
               _extendedSequenceNumber(extendedSequenceNumber)
         {
+            lockOwner();
         }
 
-        PacketT* _packet;
-        AllocatorT* _allocator;
-        transport::RtcTransport* _transport;
-        uint32_t _extendedSequenceNumber;
+        IncomingPacketAggregate(IncomingPacketAggregate&& rhs)
```
olofkallander, line 213: Nice, I will try this. Then it will move the responsibility or releaseing job count when pushing to the queue. The lie time of the IcomingAggregate will then be the life time of the ref increment.

olofkallander:
I close this as it is in PR-70

PR #73, opened 2022-04-11
Enable uplink estimation by default

RicardoMDomingues:


test/integration/IntegrationTest.cpp
```diff
@@ -793,6 +808,11 @@ void analyzeRecording(const std::vector<int16_t>& recording,
         }
     }
 
+    if (recording.size() < fftWindowSize)
```
RicardoMDomingues, line 811: The integration test were crashing because when rate controller is enabled we will receive packets on video ssrc and then the recording.size() is zero (as we only record the audio packets). The condition to exit the next `for` is `recording.size() - fftWindowSize` which underflows the the `size_t`.

I added this protection to just not get a SEGV (I thing it's preferable having a test failing than a crash)

For test to pass I ignore the analyses of packets from video ssrcs

olofkallander: Good choice.
I have not got around to do video in integration test. It requires some creation of fake VP8 headers. But it will increase code coerage a lot. But let's leave that for another PR.

olofkallander (approved):

PR #77, opened 2022-04-22
Increase code safety

RicardoMDomingues:


olofkallander (commented):
We do not need cast in JObQueue.h either, right?

olofkallander (approved):

PR #81, opened 2022-05-07
pkt log

olofkallander:


RicardoMDomingues (approved):
LGTM. Just a question, how this can affect C9 using rewrite as we will have the double of packets per second?
https://github.com/finos/SymphonyMediaBridge/blob/08004f756448f5cf76777b6c5edac1618457fca2/bridge/engine/AudioForwarderRewriteAndSendJob.cpp#L33

PR #88, opened 2022-05-24
Suggestion: Support for configurable payload type ids and rtp header extension ids

marcusspangenberg:
Use case is when the media server is the one producing the initial SDP answer and therefore cannot decide the dynamic payload type ids or rtp header extension ids. The information is already existing in the api, this PR only uses that existing data to rewrite packet headers for each receiver.

For the regular offer answer procedure nothing is changed, the default ids are used as before.

This is a draft for now, because I need to validate it some more. But I wanted to hear some feedback as well.


bridge/RtpMap.h
```diff
@@ -45,11 +67,13 @@ struct RtpMap
...
 
     Format _format;
-    uint32_t _payloadType;
+    uint16_t _payloadType;
     uint32_t _sampleRate;
     utils::Optional<uint32_t> _channels;
     std::unordered_map<std::string, std::string> _parameters;
     std::vector<std::pair<std::string, utils::Optional<std::string>>> _rtcpFeedbacks;
+    utils::Optional<uint8_t> _audioLevelExtId;
+    utils::Optional<uint8_t> _absSendTimeExtId;
```
marcusspangenberg, line 76: Extension ids are not part of the RtpMap

api/Parser.cpp
```diff
@@ -87,7 +87,7 @@ api::AllocateEndpoint::Transport parseAllocateEndpointTransport(const nlohmann::
     api::AllocateEndpoint::Transport transport;
     setIfExists(transport._ice, data, "ice");
     setIfExists(transport._iceControlling, data, "ice-controlling");
-    setIfExists(transport._dtls, data, "ice");
+    setIfExists(transport._dtls, data, "dtls");
```
olofkallander, line 90: I think we may be suffering from this bug already.

bridge/engine/AudioForwarderRewriteAndSendJob.cpp
```diff
@@ -1,19 +1,54 @@
...
+    {
+        return;
+    }
+
+    for (auto& rtpHeaderExtension : headerExtensions->extensions())
+    {
+        if (senderInboundContext._rtpMap._audioLevelExtId.isSet() &&
+            rtpHeaderExtension.getId() == senderInboundContext._rtpMap._audioLevelExtId.get())
+        {
+            rtpHeaderExtension.setId(senderInboundContext._rtpMap._audioLevelExtId.get());
```
olofkallander, line 29: We want to set it to outbound context audio level ext id, right?

bridge/engine/VideoForwarderRewriteAndSendJob.cpp
```diff
@@ -6,6 +6,34 @@
...
+    {
+        return;
+    }
+
+    for (auto& rtpHeaderExtension : headerExtensions->extensions())
+    {
+        if (senderInboundContext._rtpMap._absSendTimeExtId.isSet() &&
+            rtpHeaderExtension.getId() == senderInboundContext._rtpMap._absSendTimeExtId.get())
+        {
+            rtpHeaderExtension.setId(senderInboundContext._rtpMap._absSendTimeExtId.get());
```
olofkallander, line 29: outboundcontext rtpmap ?

bridge/engine/AudioForwarderRewriteAndSendJob.cpp
marcusspangenberg: Of course, thanks!

bridge/engine/EngineMixer.cpp
```diff
@@ -1837,6 +1836,17 @@ void EngineMixer::processIncomingRtpPackets(const uint64_t timestamp)
     {
         ++numRtpPackets;
 
+        const auto rtpHeader = rtp::RtpHeader::fromPacket(*packetInfo.packet());
+        if (!rtpHeader)
+        {
+            continue;
+        }
+        auto inboundSsrcContext = getInboundSsrcContext(rtpHeader->ssrc.get());
```
RicardoMDomingues, line 1844: Just a suggestion. Maybe we should store `inboundSsrcContext` inside of `IncomingPacketInfo` instead of `transport`. Seems like the jobs that will enqueue the `IncomingPacketInfo` already have a reference to `inboundSsrcContext` (exception is maybe the RTCP packets but we can obtain the inboundSsrc before enqueue) and `inboundSsrcContext` already have a pointer to to `transport`. In this way we avoid `getInboundSsrcContext` twice (before and after enqueue) and avoid this pattern of getting and checking if it is not a `nullptr` in almost `process` methods

olofkallander:
General comment: I think this is a really good idea. With reference to Ricardo's comment above, I think we should review how much more lookups we introduce, since the rtpmap rewrite will naturally entail some more cross referencing. If we can find some easy reductions in lookups.

marcusspangenberg:
Updated with packetInfo suggestion. I validated the rewritten payload types and extension ids in wireshark. Changing this to a proper PR.

marcusspangenberg:
Ok to merge @RicardoMDomingues @olofkallander ?

bridge/engine/EngineMixer.cpp
```diff
@@ -283,7 +283,9 @@ void EngineMixer::addVideoStream(EngineVideoStream* engineVideoStream)
         engineVideoStream->_transport.getLoggableId().c_str(),
         endpointIdHash);
 
-    auto* outboundContext = obtainOutboundSsrcContext(*engineVideoStream, engineVideoStream->_localSsrc);
+    auto* outboundContext = obtainOutboundSsrcContext(*engineVideoStream,
+        engineVideoStream->_localSsrc,
+        engineVideoStream->_feedbackRtpMap);
     if (outboundContext)
     {
         engineVideoStream->_transport.getJobQueue().addJob<SetRtxProbeSourceJob>(engineVideoStream->_transport,
```
olofkallander, line 291: It is actually in this setRtxProbe operation we should set also the rtpmap for rtx payload type. This should be set according to the SDP sent to the client.

transport/TransportImpl.h
```diff
@@ -163,6 +163,8 @@ class TransportImpl : public RtcTransport,
 
     void setAudioPayloadType(uint8_t payloadType, uint32_t rtpFrequency) override;
     void setAbsSendTimeExtensionId(uint8_t extensionId) override;
+    void setVideoRtxPayloadType(const uint16_t payloadType) override;
```
olofkallander, line 166: Remove I think.
This should be set along with the setRtxProbeSsrc.

bridge/engine/AudioForwarderRewriteAndSendJob.cpp
```diff
@@ -1,19 +1,54 @@
...
+    {
+        return;
+    }
+
+    for (auto& rtpHeaderExtension : headerExtensions->extensions())
+    {
+        if (senderInboundContext._rtpMap._audioLevelExtId.isSet() &&
+            rtpHeaderExtension.getId() == senderInboundContext._rtpMap._audioLevelExtId.get())
+        {
+            rtpHeaderExtension.setId(receiverOutboundContext._rtpMap._audioLevelExtId.get());
```
olofkallander, line 29: Now we are not checking that the outboundcontext really has a rtp hdr ext id set.
I wonder what happens with ext hdrs SMB do not care about. They will pass through unaltered and it works as long as all the clients have the same idea about the mapping. We have clients that offer other ext.

bridge/Mixer.cpp
```diff
@@ -1111,10 +1108,14 @@ bool Mixer::configureVideoStream(const std::string& endpointId,
         videoStream->_secondarySimulcastStream = secondarySimulcastStream;
     }
 
-    if (absSendTimeExtensionId.isSet())
+    videoStream->_rtpMap._absSendTimeExtId = absSendTimeExtensionId;
+    if (videoStream->_rtpMap._absSendTimeExtId.isSet())
+    {
+        videoStream->_transport->setAbsSendTimeExtensionId(videoStream->_rtpMap._absSendTimeExtId.get());
```
olofkallander, line 1114: We should post a single Job to set rtxprobeSsrc, rtxpayloadtype, and absendtimeext hdr id

marcusspangenberg: I updated the PR with setting the payload type together with the RTX probe source, but I'm leaving the abssendtime as out of scope of this PR since it also affects audio only endpoints and therefore requires some more thought.

bridge/engine/AudioForwarderRewriteAndSendJob.cpp
marcusspangenberg: True, this is the same behavior as before though. A server receiving an offer should filter out all unsupported extensions in the answer. A server sending an offer will only send the supported ones.

bridge/engine/EngineMixer.cpp
```diff
@@ -93,20 +93,24 @@ class RemoveSrtpSsrcJob : public jobmanager::CountedJob
 class SetRtxProbeSourceJob : public jobmanager::CountedJob
 {
 public:
-    SetRtxProbeSourceJob(transport::RtcTransport& transport, uint32_t ssrc, uint32_t* sequenceCounter)
+    SetRtxProbeSourceJob(transport::RtcTransport& transport,
+        const uint32_t ssrc,
+        uint32_t* sequenceCounter,
+        const uint16_t payloadType)
```
RicardoMDomingues, line 99: You forgot to set `_payloadType`

olofkallander:
Except for th _payloadType, perhaps we should merge this and create another ticket to refine the rtx rtpmap setting and handling of unsupported hdr extensions. SMB may receive hdr extensions SMB does not support. The unsupported will be expected to pass through as specified in RFCs also. This cannot be done if the mapping collides with the receiver mapping. In those cases the extensions will not be remapped because we only remap those that SMB understands.

bridge/engine/EngineMixer.cpp
marcusspangenberg: Sorry about that, should be fixed now.

RicardoMDomingues (approved):
LGTM

PR #91, opened 2022-05-30
RTC-12879 audio lastN

olofkallander:
Limit alstN audio as config. Reduces number of audio sinks in client.


bridge/engine/ActiveMediaList.cpp
```diff
@@ -89,11 +91,10 @@ bool ActiveMediaList::addAudioParticipant(const size_t endpointIdHash)
         _dominantSpeaker = endpointIdHash;
     }
 
-    if (_audioSsrcRewriteMap.size() == _maxActiveListSize)
+    if (_audioSsrcRewriteMap.size() == _audioLastN)
```
olofkallander, line 94: Why do we not just add audio sources until we run out of audio ssrcs. Why stop at "top3"

RicardoMDomingues (approved):

PR #95, opened 2022-06-02
Parse relay candidates

RicardoMDomingues:


RicardoMDomingues (commented):
https://github.com/finos/SymphonyMediaBridge/blob/e061de217876a9dcad816817f128da2782b535d4/bridge/ApiRequestHandler.cpp#L83-L86

I didn't add the translation for RELAY candidates here as smb never produce RELAY candidates.

olofkallander (approved):

PR #96, opened 2022-06-07
Do not present more participants than lastN on ssrcRewriteMap

RicardoMDomingues:


bridge/engine/ActiveMediaList.cpp
```diff
@@ -86,6 +86,12 @@ bool ActiveMediaList::addAudioParticipant(const size_t endpointIdHash)
     }
 
     _audioParticipants.emplace(endpointIdHash, AudioParticipant());
+
+    if (_audioSsrcRewriteMap.size() == _maxActiveListSize)
```
olofkallander, line 90: maxActiveListSize is about 10.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -94,6 +100,7 @@ bool ActiveMediaList::addAudioParticipant(const size_t endpointIdHash)
     uint32_t ssrc;
     if (!_audioSsrcs.pop(ssrc))
```
olofkallander, line 101: Using the existing audioSsrcs, which are audio.lastN+2, 5 as default, should not be a problem.

RicardoMDomingues:
This PR does not fix the issue detected on e2e tests

PR #98, opened 2022-06-08
Olofk/fixdominantspeaker

olofkallander:


bridge/engine/ActiveMediaList.cpp
```diff
@@ -288,8 +288,13 @@ size_t ActiveMediaList::rankSpeakers(float& currentDominantSpeakerScore)
...
         const auto& participantLevels = participantLevelEntry.second;
-        const float participantScore = participantLevels._maxRecentLevel - participantLevels._noiseLevel;
-        // std::max(0.0f, participantLevels._maxRecentLevel - participantLevels._noiseLevel);
+        if (participantLevels._maxRecentLevel == 0)
+        {
+            continue;
+        }
+
+        const float participantScore =
+            std::max(0.0f, participantLevels._maxRecentLevel - participantLevels._noiseLevel);
```
RicardoMDomingues, line 297: Don't we want it like this:

```suggestion
      const float participantScore =
            std::max(0.0f, participantLevels._maxRecentLevel - participantLevels._noiseLevel);
            
        if (participantScore == 0)
        {
            continue;
        }
```

PR #99, opened 2022-06-24
Log nominated ICE candidate

RicardoMDomingues:


transport/TransportImpl.cpp
```diff
@@ -2024,6 +2024,11 @@ void TransportImpl::onIcePreliminary(ice::IceSession* session,
             sourcePort.toString().c_str());
         _selectedRtp = static_cast<transport::Endpoint*>(endpoint); // temporary selection
         _peerRtpPort = sourcePort;
+        logger::info("temporary candidate selected %s %s",
```
olofkallander, line 2027: Should we perhaps repalce the log above with this and just add some more params to the log?
Seems unnecessary to log both

olofkallander (approved):

transport/TransportImpl.cpp
RicardoMDomingues: Yes, I thought the same but as it other is debug level I let both as they log different info. But I will merge both logs parms in only one

PR #100, opened 2022-06-28
Turn off Bandwidth Estimation for audio-only calls

reddvl1980:
Bandwidth Estimator (BE) uses RTP/RTCP stress loading by sending extra "padding" in order to model the network condition and select the best possible configurations for video conferences. It makes no sense for audio-only calls, where it just consumes traffic, which negatively affect clients on metered connections.

- refactor pacing buffer draining;
- disable RTCP-padding for audio calls not using SSRC-rewriting;
- disable RTP-padding for video calls when video source becomes idle.

### DIFF
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -320,6 +320,14 @@ void EngineMixer::addVideoStream(EngineVideoStream* engineVideoStream)
     sendVideoStreamToRecording(*engineVideoStream, true);
 }
 
+void stopProbingVideoStream(EngineVideoStream* engineVideoStream) 
+{
+    engineVideoStream->_transport.getJobQueue().addJob<SetRtxProbeSourceJob>(engineVideoStream->_transport,
+                                                engineVideoStream->_localSsrc,
+                                                nullptr,
+                                                engineVideoStream->_feedbackRtpMap._payloadType);
+}
+
 void EngineMixer::removeVideoStream(EngineVideoStream* engineVideoStream)
 {
     engineVideoStream->_transport.getJobQueue().getJobManager().abortTimedJob(engineVideoStream->_transport.getId(),
@@ -331,10 +339,7 @@ void EngineMixer::removeVideoStream(EngineVideoStream* engineVideoStream)
         outboundContext->_markedForDeletion = true;
         if (engineVideoStream->_transport.isConnected())
         {
-            engineVideoStream->_transport.getJobQueue().addJob<SetRtxProbeSourceJob>(engineVideoStream->_transport,
-                engineVideoStream->_localSsrc,
-                nullptr,
-                engineVideoStream->_feedbackRtpMap._payloadType);
+            stopProbingVideoStream(engineVideoStream);
             auto goodByePacket = createGoodBye(outboundContext->_ssrc, outboundContext->_allocator);
             if (goodByePacket)
             {
@@ -1065,6 +1070,11 @@ void EngineMixer::checkPacketCounters(const uint64_t timestamp)
                     message._command.freeVideoPacketCache._ssrc = outboundContextEntry.first;
                     message._command.freeVideoPacketCache._endpointIdHash = videoStreamEntry.first;
                     _messageListener.onMessage(std::move(message));
+                }
+
+                // Removing BE probing for this video stream SSRC.
+                {
+                     stopProbingVideoStream(videoStreamEntry.second);
                 }
 
                 logger::info("Removing idle outbound context ssrc %u, endpointIdHash %lu",
--- a/transport/TransportImpl.cpp
+++ b/transport/TransportImpl.cpp
@@ -2357,24 +2357,11 @@ void TransportImpl::runTick(uint64_t timestamp)
 
 void TransportImpl::doRunTick(const uint64_t timestamp)
 {
-    if (!_config.rctl.enable || !_config.bwe.useUplinkEstimate)
-    {
-        while (!_rtxPacingQueue.empty())
-        {
-            protectAndSendRtp(timestamp, _rtxPacingQueue.fetchBack());
-        }
-        while (!_pacingQueue.empty())
-        {
-            protectAndSendRtp(timestamp, _pacingQueue.fetchBack());
-        }
-        _pacingInUse = false;
-        return;
+    bool useBE = _config.rctl.enable && _config.bwe.useUplinkEstimate;
+    drainPacingBuffer(timestamp, useBE);
+    if (useBE) {
+        sendPadding(timestamp);
     }
-
-    drainPacingBuffer(timestamp);
-
-    sendPadding(timestamp);
-
     _pacingInUse = !_pacingQueue.empty() || !_rtxPacingQueue.empty();
 }
 
@@ -2384,13 +2371,13 @@ inline memory::UniquePacket TransportImpl::tryFetchPriorityPacket(size_t budget)
     return !queue.empty() && budget >= queue.back()->getLength() + _config.ipOverhead ? queue.fetchBack() : nullptr;
 }
 
-void TransportImpl::drainPacingBuffer(uint64_t timestamp) 
+void TransportImpl::drainPacingBuffer(uint64_t timestamp, bool useEstimatedBudget) 
 {
-    auto budget = _rateController.getPacingBudget(timestamp);
+    auto budget = useEstimatedBudget ? _rateController.getPacingBudget(timestamp) : SIZE_MAX;
     while (auto packet = tryFetchPriorityPacket(budget)) {
         budget -= packet->getLength() + _config.ipOverhead;
         protectAndSendRtp(timestamp, std::move(packet));
-    } 
+    }
 }
 
 } // namespace transport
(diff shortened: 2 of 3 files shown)


transport/TransportImpl.cpp
```diff
@@ -2390,44 +2357,27 @@ void TransportImpl::runTick(uint64_t timestamp)
...
 
-    auto budget = _rateController.getPacingBudget(timestamp);
-    while (!_pacingQueue.empty() || !_rtxPacingQueue.empty())
-    {
-        auto* pacingQueue = &_pacingQueue;
-        if (!_rtxPacingQueue.empty())
-        {
-            pacingQueue = &_rtxPacingQueue;
-        }
+inline memory::UniquePacket TransportImpl::tryFetchPriorityPacket(size_t budget)
```
RicardoMDomingues, line 2368: [minor] `inline` is not make any difference on this case on modern c++ compilers. Compiler in the end will decided inline it or not depend what it think is the best decision. In my opinion we should try to keep code as less verbose as possible and use `inline` when it is required

transport/TransportImpl.h
```diff
@@ -301,6 +303,8 @@ class TransportImpl : public RtcTransport,
     RtpSenderState& getOutboundSsrc(uint32_t ssrc, uint32_t rtpFrequency);
 
     void onTransportConnected();
+    void drainPacingBuffer(uint64_t timestamp, bool useEstimatedBudget = true);
```
RicardoMDomingues, line 306: [minor] personally I am not fan of default values that can change how the method behaves. I mean, I think it is useful for things that will not affect the method output (for example: use a default value for a buffer size and add possibility to give a better value that optimizes a specific use case), otherwise seems a bit error prone

bridge/engine/EngineMixer.cpp
```diff
@@ -1067,6 +1072,11 @@ void EngineMixer::checkPacketCounters(const uint64_t timestamp)
                     _messageListener.onMessage(std::move(message));
                 }
 
+                // Removing BE probing for this video stream SSRC.
+                {
+                     stopProbingVideoStream(videoStreamEntry.second);
```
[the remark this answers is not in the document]
RicardoMDomingues: Also we have to start probing again in the case we create new outboundcontext for a videostream with a stopped probe

transport/TransportImpl.h
```diff
@@ -41,6 +41,9 @@ struct SctpConfig;
 
 namespace transport
 {
+
+typedef memory::RandomAccessBacklog<memory::UniquePacket, 512> pacing_queue_t;
```
RicardoMDomingues, line 45: I like this. but I guess there is a kind of a rule in this project to not create alias for types like this. This is because it creates an indirection for who is reading the code I guess. But the team has changed so maybe we can discuss with @olofkallander if we can change some code rules . My vote is yes on this one :)

transport/TransportImpl.h
reddvl1980: I am totally agree with you on this one.  I checked the code prior to changes, to see whether default values are in use and found in the same TransportImpl.h (sendReports(..., bool rembReady = false)), so decided not to bring new rules into the project :) 

Changing to what I personally consider the best: non-default enum parameter that explicitly (in English) sets the mode.

transport/TransportImpl.h
reddvl1980: Let's keep it this way then. When Olof is back, we'll talk.

transport/TransportImpl.cpp
reddvl1980: Agree, since it's not "free function included from multiple files" it is unnecessary since. Removed inline spec.

RicardoMDomingues (commented):
General comment about code style: 
If you are use vscode I advise you to install extension 
`xaver.clang-format` and add the following setting in your `.vscode/settings.json`:
` "editor.formatOnSave" : true`

So you don't need to think about on code style. That will format the code automatically on save based on our rules in `.clang-format` file.

Hope that extension works for you, mine has broken with an update of a dependency and I couldn't fix it yet :(

transport/TransportImpl.cpp
```diff
@@ -2390,44 +2357,27 @@ void TransportImpl::runTick(uint64_t timestamp)
...
-        {
-            memory::UniquePacket packet(pacingQueue->fetchBack());
-            budget -= packet->getLength() + _config.ipOverhead;
-            protectAndSendRtp(timestamp, std::move(packet));
-        }
-        else
-        {
-            break;
-        }
+void TransportImpl::drainPacingBuffer(uint64_t timestamp, DrainPacingBufferMode mode)
```
olofkallander, line 2374: Instead f adding that enum, why don't we just pass the drain budget to the method. Either SIZE_MAX or _rateController.getPacingBudget. Seems like we add new "types" for no solid reason.

bridge/engine/EngineMixer.cpp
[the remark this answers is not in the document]
reddvl1980: Updated in the new commits:
- never toggle BE for video steams one-by-one;
- if no video packets were sent from any of the clients for the last 30 s - disable BE probing for all video streams;
- as soon as one video packet arrive from any one of the client - enable BE probing for all video streams.

bridge/engine/EngineMixer.cpp
```diff
@@ -290,16 +291,7 @@ void EngineMixer::addVideoStream(EngineVideoStream* engineVideoStream)
...
-        engineVideoStream->_localSsrc,
-        engineVideoStream->_feedbackRtpMap);
-    if (outboundContext)
-    {
-        engineVideoStream->_transport.getJobQueue().addJob<SetRtxProbeSourceJob>(engineVideoStream->_transport,
-            engineVideoStream->_localSsrc,
-            &outboundContext->_sequenceCounter,
-            outboundContext->_rtpMap._payloadType);
-    }
+    startProbingVideoStream(*engineVideoStream);
```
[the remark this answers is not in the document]
reddvl1980: Agree. Committed your suggestion.

bridge/engine/EngineMixer.cpp
```diff
@@ -3213,10 +3203,66 @@ void EngineMixer::checkVideoBandwidth(const uint64_t timestamp)
...
+
+    if (enableBEProbing == _probingVideoStreams)
+        return;
+
+    for (auto& videoStream: _engineVideoStreams) {
+        if (videoStream.second) {
+            if (enableBEProbing)
+                startProbingVideoStream(*videoStream.second);
+            else
+                stopProbingVideoStream(*videoStream.second);
```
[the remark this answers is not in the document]
reddvl1980: Fixed with clang-format.

transport/TransportImpl.cpp
```diff
@@ -2390,44 +2357,27 @@ void TransportImpl::runTick(uint64_t timestamp)
...
-    while (!_pacingQueue.empty() || !_rtxPacingQueue.empty())
-    {
-        auto* pacingQueue = &_pacingQueue;
-        if (!_rtxPacingQueue.empty())
-        {
-            pacingQueue = &_rtxPacingQueue;
-        }
+memory::UniquePacket TransportImpl::tryFetchPriorityPacket(size_t budget)
+{
+    auto queue = _rtxPacingQueue.empty() ? _pacingQueue : _rtxPacingQueue;
```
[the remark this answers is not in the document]
reddvl1980: Good catch! Fixed.

test/integration/IntegrationTest.cpp
```diff
@@ -1284,6 +1298,210 @@ TEST_F(IntegrationTest, plain)
     }
 }
 
+TEST_F(IntegrationTest, audioOnlyNoPadding)
+{
+    if (__has_feature(address_sanitizer) || __has_feature(thread_sanitizer))
```
RicardoMDomingues, line 1303: Not sure why we are ignoring the tests when address sanitizer and thread sanitazer are enabled. But it seems a common patter in all tests in `IntegrationTest`. If the reason is we don't to want to spend the time running the `IntegrationTest` in all build flavours I wold suggest to change on `cmake` level. 
Only include IntegrationTest.cpp when `${CMAKE_BUILD_TYPE}` is not `DCHECK` nor `TCHECK` so we can get rid of this boilerplate condition

RicardoMDomingues (approved):
LGTM

bridge/engine/EngineMixer.cpp
```diff
@@ -3213,10 +3203,66 @@ void EngineMixer::checkVideoBandwidth(const uint64_t timestamp)
...
+{
+    engineVideoStream._transport.getJobQueue().addJob<SetRtxProbeSourceJob>(engineVideoStream._transport,
+                                                engineVideoStream._localSsrc,
+                                                nullptr,
+                                                engineVideoStream._feedbackRtpMap._payloadType);
+}
+
+bool EngineMixer::isVideoInUse(const uint64_t timestamp, const uint64_t threshold) const
+{
+    bool videoInUse = false;
```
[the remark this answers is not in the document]
reddvl1980: Done

reddvl1980:
Closing in favor of the PR created on the "origin": https://github.com/finos/SymphonyMediaBridge/pull/102

PR #103, opened 2022-07-05
Protect api parser against access to missing json nodes

RicardoMDomingues:
We access unsafely to some json properties which can cause crash on smb when we receive requests thought the api that not contains some value that we expect.

Changes:
- Add missing protection on access to some json array properties and raise an exception (which will be translated to a `400 BAD_REQUEST`) when the property is required
- Add an anonymous function that return an empty array when optional properties are not present, removing the boilerplate code checking if the array is present before iterate it

### DIFF
--- a/api/Parser.cpp
+++ b/api/Parser.cpp
@@ -346,7 +346,7 @@ EndpointDescription parsePatchEndpoint(const nlohmann::json& data, const std::st
             videoChannel._transport.set(parsePatchEndpointTransport(videoJson["transport"]));
         }
 
-        for (const auto& ssrcJson : videoJson["ssrcs"])
+        for (const auto& ssrcJson : safeJsonArray(videoJson, "ssrcs"))
         {
             const auto ssrc = ssrcJson.is_string() ? std::stoul(ssrcJson.get<std::string>()) : ssrcJson.get<uint32_t>();
             videoChannel._ssrcs.push_back(ssrc);
--- a/api/Parser.cpp
+++ b/api/Parser.cpp
@@ -4,10 +4,10 @@
 namespace
 {
 
-const nlohmann::json EMPTY_JSON_ARRAY = nlohmann::json::array();
-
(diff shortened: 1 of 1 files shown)


reddvl1980:
I've got two questions:
- what's our policy on throwing exceptions in C++;
- do we have unit tests for the Parser. If we don't - we should. If we have - we need to add new tests for JSON arrays.

api/Parser.cpp
```diff
@@ -4,6 +4,29 @@
 namespace
 {
 
+const nlohmann::json EMPTY_JSON_ARRAY = nlohmann::json::array();
+
+const nlohmann::json& safeJsonArray(const nlohmann::json& data, const char* arrayProperty)
```
olofkallander, line 9: Can we call this optionalJsonArray for consistency with requiredJsonArray.

api/Parser.cpp
```diff
@@ -4,6 +4,29 @@
 namespace
 {
 
+const nlohmann::json EMPTY_JSON_ARRAY = nlohmann::json::array();
```
olofkallander, line 7: I do not think this alias is necessary. Or do we run into compiler problems at line 12? At line 12 it is still obvious it is an empty array.

olofkallander (changes requested):

api/Parser.cpp
RicardoMDomingues: I changed the place to be more clear the purpose of that variable. Now is a `static` variable scoped in `optionalJsonArray` function

olofkallander:
> I do not think this alias is necessary. Or do we run into compiler problems at line 12? At line 12 it is still obvious it is an empty array. 

Sorry I misread it. I realised when I got home that you probably returned a reference there.

olofkallander (approved):

reddvl1980:
Do we use std::optional? It fits the pattern parser. Drawback (or not?) would be that exceptions would be thrown after check for being null, in the body of the caller, rather than in the parser itself.

reddvl1980 (approved):

PR #104, opened 2022-07-06
refactor int test

olofkallander:
Some support code to simplify impl of integration test

### DIFF
--- a/test/integration/IntegrationTest.cpp
+++ b/test/integration/IntegrationTest.cpp
@@ -12,6 +12,7 @@
 #include "test/integration/SampleDataUtils.h"
 #include "test/integration/emulator/ApiChannel.h"
 #include "test/integration/emulator/AudioSource.h"
+#include "test/integration/emulator/HttpRequests.h"
 #include "test/integration/emulator/SfuClient.h"
 #include "transport/DataReceiver.h"
 #include "transport/RtcTransport.h"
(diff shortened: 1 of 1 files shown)


RicardoMDomingues (commented):
LGTM. Just minor comments

test/integration/IntegrationTest.h
```diff
@@ -36,3 +36,100 @@ struct IntegrationTest : public ::testing::Test
...
+                }
+            }
+
+            if (connectedCount == _clients.size())
+            {
+                return true;
+            }
+            utils::Time::nanoSleep(1 * utils::Time::sec);
+            logger::debug("waiting for connect...", "test");
+        }
```
RicardoMDomingues, line 91: I think this can be simplified, like:
```
while (utils::Time::getAbsoluteTime() - start < timeout)
{
    auto it = std::find_if_not(_clients.begin(), _clients.end(), 
        [] (auto c) { return c->_transport->isConnected(); });
        
    if (it == _clients.end())
    {
         return true;
    }
    utils::Time::nanoSleep(1 * utils::Time::sec);
    logger::debug("waiting for connect...", "test");
}
```

test/integration/IntegrationTest.h
```diff
@@ -36,3 +36,100 @@ struct IntegrationTest : public ::testing::Test
 
     void initBridge(config::Config& config);
 };
+
+template <typename T>
+class GroupCall
+{
+public:
+    GroupCall(const std::vector<T*> clients) : _clients(clients) {}
```
RicardoMDomingues, line 44:
```suggestion
    GroupCall(const std::vector<T*>& clients) : _clients(clients) {}    
```
--

I know this is a test code only but maybe it would be even better use `std::initializer_list` to avoid the temporary heap allocation created by the temporary `std::vector`. like:
```suggestion
    GroupCall(std::initializer_list<T*> clients) : _clients(clients) {}
```
And you give the braced initialization, like:
`GroupCall<SfuClient<ColibriChannel>> groupCall = {&client1, &client2, &client3};`

RicardoMDomingues (approved):

test/integration/IntegrationTest.cpp
```diff
@@ -1338,35 +1341,19 @@ TEST_F(IntegrationTest, audioOnlyNoPadding)
...
     client3.initiateCall(baseUrl, conf.getId(), false, true, false, false);
 
-    EXPECT_TRUE(client1._channel.isSuccess());
-    EXPECT_TRUE(client2._channel.isSuccess());
-    EXPECT_TRUE(client3._channel.isSuccess());
-
-    if (!client1._channel.isSuccess() && client2._channel.isSuccess() && client3._channel.isSuccess())
+    if (!groupCall.connect(utils::Time::sec * 8))
     {
+        EXPECT_TRUE(false);
```
RicardoMDomingues, line 1353: Makes the intention clear and gives a better message about the failing when test fail because of thos
```suggestion
        FAIL() << "Clients not connected on time";
```

RicardoMDomingues (commented):
LGTM @reddvl1980.

I have only a general comment about IntegrationTest. This started being something small and it is getting bigger (which is good, means we are doing more completed tests) but put everything on same *.cpp it does not looks pretty for this amount of code/classes
 
My suggestion (doesn't need to be in this PR) is create files for this classes and let IntegrationTest.cpp for the test definition only.

Also with the increase of test scenarios we can even think about in make multiple integration test files (like we do for the unit test) and create an own executable for integration test (outside of unit tests executable)

test/integration/IntegrationTest.cpp
```diff
@@ -272,11 +345,11 @@ class Channel
         }
     }
 
-    void sendResponse(const std::pair<std::string, std::string>& iceCredentials,
+    virtual void sendResponse(const std::pair<std::string, std::string>& iceCredentials,
```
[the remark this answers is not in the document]
reddvl1980: I agree. Let's refactor it (split) a new PR to keep branches short-living.

reddvl1980 (approved):

RicardoMDomingues (approved):

PR #105, opened 2022-07-08
RTC-13082 report SMB API capability

reddvl1980:
Add /about/capabilities endpoint that reports [SMB_API] now (when barbelling is ready, it'll report [SMB_API, BARBELLING].

### DIFF
--- a/bridge/ApiRequestHandler.cpp
+++ b/bridge/ApiRequestHandler.cpp
@@ -589,7 +589,7 @@ httpd::Response ApiRequestHandler::handleAbout(const httpd::Request& request,
     }
     else if (utils::StringTokenizer::isEqual(nextToken, "capabilities"))
     {
-        httpd::Response response(httpd::StatusCode::OK, "[SMB_API]");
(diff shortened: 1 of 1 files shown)


bridge/ApiRequestHandler.cpp
```diff
@@ -588,6 +587,12 @@ httpd::Response ApiRequestHandler::handleAbout(const httpd::Request& request,
         response._headers["Content-type"] = "text/json";
         return response;
     }
+    else if (utils::StringTokenizer::isEqual(nextToken, "capabilities"))
+    {
+        httpd::Response response(httpd::StatusCode::OK, "[\"SMB_API.1\", \"SSRC_REWRITING.1\"]");
```
olofkallander, line 592: Please use camel case for these capabilities in the response body.
This just makes it harder to read and eyes ache.

reddvl1980: As agreed with @olofkallander , I made it lowercase in SFU, but with "_" to facilitate robust parsing in CS.

olofkallander (approved):

PR #106, opened 2022-07-13
refactor endpoint life cycle

olofkallander:
RTC-13118


utils/ScopedIncrement.h
```diff
@@ -10,12 +11,20 @@ class ScopedIncrement
...
+    {
+#if DEBUG
+        auto value = --_counter;
+        assert(value != 0xFFFFFFFFu);
+#else
+        --counter;
+#endif
+    }
 
     std::atomic_uint32_t& getCounter() { return _counter; }
```
reddvl1980, line 24: This looks fishy since it allows to modify _counter. I have not found usage of it. If what is needed is just a check of counter's value  - consider returning "const uint32" and mark method as "const".

transport/UdpEndpoint.cpp
```diff
@@ -26,21 +28,22 @@ class UnRegisterListenerJob : public jobmanager::Job
     Endpoint::IEvents* _listener;
 };
 
-class UnRegisterStunListenerJob : public jobmanager::Job
+class UnRegisterNotifyListenerJob : public jobmanager::Job
 {
 public:
-    UnRegisterStunListenerJob(UdpEndpoint& endpoint, __uint128_t transactionId)
+    UnRegisterNotifyListenerJob(UdpEndpoint& endpoint, Endpoint::IEvents& listener)
```
reddvl1980, line 34: Neatpicking: pass parameters as "const Type&". Though here methods are tiny and it does not worth changing. But I noticied a pattern.

transport/TransportImpl.h
```diff
@@ -61,8 +63,8 @@ class TransportImpl : public RtcTransport,
         const ice::IceRole iceRole,
         const bwe::Config& bweConfig,
         const bwe::RateControllerConfig& rateControllerConfig,
-        const std::vector<Endpoint*>& rtpEndPoints,
-        const std::vector<ServerEndpoint*>& tcpEndPoints,
+        const Endpoints& rtpEndPoints,
+        const ServerEndpoints& tcpEndPoints,
         TcpEndpointFactory* tcpEndpointFactory,
```
reddvl1980, line 68: Can it be passed/owned as shared_ptr instead of raw ptr?

utils/ScopedIncrement.h
reddvl1980: Disregard. I got what you wanted to do (read a comment in job.h)

transport/TcpServerEndpoint.h
```diff
@@ -24,18 +25,19 @@ class TcpServerEndpoint : public ServerEndpoint, public RtcePoll::IEventListener
         RtcePoll& rtcePoll,
         size_t maxPengingSessions,
         TcpServerEndpoint::IEvents* defaultListener,
+        TcpEndpointFactory& tcpEndpointFactory,
```
reddvl1980, line 28: See similar comment: Can it be passed/owned as shared_ptr instead of raw ptr? Why passing differently?

transport/TcpServerEndpoint.h
```diff
@@ -46,7 +48,7 @@ class TcpServerEndpoint : public ServerEndpoint, public RtcePoll::IEventListener
 
 public: // internal job methods
```
reddvl1980, line 49: Why internal job's method are public?

transport/RecordingTransport.cpp
```diff
@@ -125,10 +125,10 @@ void RecordingTransport::protectAndSend(memory::UniquePacket packet)
         return;
     }
 
-    protectAndSend(std::move(packet), _peerPort, _recordingEndpoint);
+    protectAndSend(std::move(packet), _peerPort, *_recordingEndpoint);
```
reddvl1980, line 128: The last parameter for this calling this method is not needed.

olofkallander:
> Disregard. I got what you wanted to do (read a comment in job.h) 

I think you are right. It is a weird construct and when I look at how it is used it is in test to pass a param into a method that does nothing.

transport/TcpServerEndpoint.h
reddvl1980: I see. I would definitely prefer combining all job-called "internal" methods into an interface, and pass it to the job.

Pros: clearly readable code and intent. Speed does not change. Also, combining them into interfaces would make more prominent similarities (and allow to combine and thus reduce codebase size).
Cons: codebase size slightly increases.

transport/RtcePoll.cpp
```diff
@@ -208,21 +221,17 @@ void RtcePollImpl::run()
 
     while (_running)
     {
```
olofkallander, line 223: The wait on kernel_fd will not immediately return if monitored sockets is empty. Therefore, there is no need for sleep to avoid tight loop in case the socket list is empty.

PR #107, opened 2022-07-15
Use TrackerWithSnapshot for BaseUdpEndpoint metrics

reddvl1980:


utils/Trackers.h
```diff
@@ -133,6 +133,26 @@ class RateTracker
     uint64_t _prevTimestamp;
 };
 
+template <size_t NUM_BUCKETS, uint64_t BUCKET_SIZE_MS, uint64_t SNAPSHOT_DELTA_MS>
+struct TrackerWithSnapshot
+{
+    TrackerWithSnapshot() : tracker(BUCKET_SIZE_MS), _lastSnapshotUpdatedAt(0) { snapshot.store(0.0); }
+    utils::RateTracker<NUM_BUCKETS> tracker;
+    std::atomic<double> snapshot;
```
olofkallander, line 141: there is a typedef for this std::atomic_double

transport/BaseUdpEndpoint.h
```diff
@@ -71,6 +71,18 @@ class BaseUdpEndpoint : public Endpoint, RtcePoll::IEventListener
         memory::UniquePacket packet;
     };
 
+    struct RateMetrics
+    {
+        utils::TrackerWithSnapshot<10, utils::Time::ms * 100, utils::Time::sec> receiveTracker;
+        utils::TrackerWithSnapshot<10, utils::Time::ms * 100, utils::Time::sec> sendTracker;
+        EndpointMetrics toEndpointMetrics(size_t size) const
+        {
+            return EndpointMetrics(size,
```
olofkallander, line 80: should we rename it queueSize to be more specific what is stored in EndpointMetrics.

olofkallander:
I like it. One rename perhaps.

utils/Trackers.h
reddvl1980: Are you sure?

reddvl1980: I mean atomic<double> is not fully implemented, and I can't find atomic_double macro.

transport/BaseUdpEndpoint.h
reddvl1980: Done.

PR #113, opened 2022-07-20
Refactor ApiRequestHandler

reddvl1980:
- split megamethod into "small" actions
- extract helpers into bridge/actions/utils.h
- use map<actionType, action> to invoke handlers


bridge/actions/ActionContext.h
```diff
@@ -0,0 +1,31 @@
...
+{
+struct ActionContext
+{
+    ActionContext(bridge::MixerManager& mixerManager, transport::SslDtls& sslDtls)
+        : _mixerManager(mixerManager),
+          _sslDtls(sslDtls),
+          _lastAutoRequestId(0)
+    {
+    }
+    bridge::MixerManager& _mixerManager;
```
olofkallander, line 27: while refactoring. Should we expose these without _.  As they are not private to all the functions operating on this struct?

bridge/actions/ActionContext.h
```diff
@@ -0,0 +1,31 @@
...
+    transport::SslDtls& _sslDtls;
+    std::atomic<uint32_t> _lastAutoRequestId;
```
olofkallander, line 29: Why is this part of ActionContext when it is only used by ApiRequestHandler to init RequestLogger.
The name consists of three generic words, "last", "auto", "request". Does not tell reader much about what this is. It seems to be a transaction id for logging.

bridge/actions/Stats.cpp
```diff
@@ -0,0 +1,18 @@
+#include "ApiActions.h"
```
olofkallander, line 1: Can we rename this file. We have too many Stats.cpp. The others are named Get....

bridge/actions/About.cpp
```diff
@@ -0,0 +1,36 @@
+#include "ApiActions.h"
```
olofkallander, line 1: actions folder name is a bit generic as well. 
These are mainly about http endpoint implementations, right?

bridge/ApiRequestHandler.cpp
```diff
@@ -40,358 +18,67 @@ httpd::Response makeErrorResponse(httpd::StatusCode statusCode, const std::strin
...
-            utils::format("rtp payload '%s' not supported", payloadType._name.c_str()));
-    }
-
-    for (const auto& parameter : payloadType._parameters)
-    {
-        rtpMap._parameters.emplace(parameter.first, parameter.second);
-    }
-
-    return rtpMap;
+    _actionMap.emplace(ApiActions::ABOUT, ::bridge::handleAbout);
```
olofkallander, line 33: I wonder. One way to put it is that this looks like C-implementation boiler plate of virtual methods.
Since you already added multiple inheritance to ApiRequestHandler, one could have put all the requesthandlers in super classes that one inherits from ApiRequestHandler and call them directly form onRequest. That way you would have type check, that method is implemented, at compile time. I am not that fond of this interpretation of request URL into enum into "switch/map" into function pointers to implementation.

bridge/actions/utils.cpp
```diff
@@ -0,0 +1,366 @@
+#include "utils.h"
```
olofkallander, line 1: Is this file more like ApiHelpers.cpp

olofkallander (changes requested):
I like that we split the http endpoint handlers into separate files. But I think it could be done without creating a runtime map to actions. Inheriting the handler implementation classes is one option.

bridge/ApiRequestHandler.cpp
```diff
@@ -40,358 +18,67 @@ httpd::Response makeErrorResponse(httpd::StatusCode statusCode, const std::strin
...
-
-                logger::debug("Add simulcast level main ssrc %u feedback ssrc %u, content %s, endpointId %s",
-                    "ApiRequestHandler",
-                    simulcastSsrc,
-                    feedbackGroup->_ssrcs[1],
-                    toString(simulcastStream._contentType),
-                    endpointId.c_str());
+                auto nextToken = utils::StringTokenizer::tokenize(outToken, '/');
+                if (nextToken.next)
+                    return ApiActions::GET_ENDPOINT_INFO;
```
olofkallander, line 60: We could have just made the call to the actionHandler here. I do not see it would make the structure more complicated.

bridge/actions/ActionContext.h
reddvl1980: Fixed.

bridge/actions/ActionContext.h
reddvl1980: Removed from this context.

bridge/ApiRequestHandler.cpp
reddvl1980: Agree!

bridge/actions/About.cpp
reddvl1980: Renamed to endpointActions.

olofkallander (approved):

PR #114, opened 2022-07-21
RTC-13095 implement active talker detection

reddvl1980:


bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -132,10 +132,21 @@ void AudioForwarderReceiveJob::run()
     const auto rtpHeaderExtensions = rtpHeader->getExtensionHeader();
     if (rtpHeaderExtensions)
     {
+        auto c9infoExtId = _ssrcContext._rtpMap._c9infoExtId.isSet() ? _ssrcContext._rtpMap._c9infoExtId.get() : 0;
+
         int32_t audioLevel = -1;
+        utils::Optional<bool> is_ptt;
```
olofkallander, line 138: We should stick to isPtt;

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -132,10 +132,21 @@ void AudioForwarderReceiveJob::run()
...
 
+        // Use one loop to check both extensions for speed up.
```
olofkallander, line 140: Remove comments that are not extremely necessary to explain what is happening. Code should speak for itself.

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -147,6 +158,7 @@ void AudioForwarderReceiveJob::run()
         }
 
         _activeMediaList.onNewAudioLevel(_sender->getEndpointIdHash(), audioLevel);
+        _activeMediaList.onNewPtt(_sender->getEndpointIdHash(), is_ptt);
 
         if (audioLevel >= _silenceThresholdLevel)
```
olofkallander, line 163: Is this really correct? The silenceThreshold is 127. Iirc the audioLevel in packet has omitted sign, so level is 1 - pktLevel.
So >= is correct to detect it is now completely silent. But the default value if there is no audio level in packet is -1. Why not 0 then.
I think even the ext.data[0] is unsigned.

bridge/engine/ActiveMediaList.h
```diff
@@ -160,6 +163,7 @@ class ActiveMediaList
         int32_t _nonZeroLevelsShortWindow;
         float _maxRecentLevel;
         float _noiseLevel;
+        bool _is_ptt;
```
olofkallander, line 166: isPtt

bridge/engine/ActiveMediaList.cpp
```diff
@@ -307,6 +309,19 @@ size_t ActiveMediaList::rankSpeakers(float& currentDominantSpeakerScore)
     return speakerCount;
 }
 
+void ActiveMediaList::onNewPtt(const size_t endpointIdHash, utils::Optional<bool> is_ptt)
```
olofkallander, line 312: Why would we call this method if we do not have an ext id?
Why make it Optional type

bridge/engine/ActiveMediaList.h
```diff
@@ -223,6 +227,7 @@ class ActiveMediaList
     void updateLevels(const uint64_t timestampMs);
     void updateActiveAudioList(size_t endpointIdHash);
     bool updateActiveVideoList(const size_t endpointIdHash);
+    std::atomic_bool _c9_conference;
```
olofkallander, line 230: We do not need this do we?

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -147,6 +158,7 @@ void AudioForwarderReceiveJob::run()
         }
 
         _activeMediaList.onNewAudioLevel(_sender->getEndpointIdHash(), audioLevel);
+        _activeMediaList.onNewPtt(_sender->getEndpointIdHash(), is_ptt);
```
olofkallander, line 161: Do not make the call if we do not have header.

bridge/engine/AudioForwarderReceiveJob.cpp
reddvl1980: Fixed.

bridge/engine/AudioForwarderReceiveJob.cpp
reddvl1980: Removed.

bridge/engine/ActiveMediaList.cpp
reddvl1980: Agree. I was think like "it does not metter where we check Optional, so why not to defer it to the very end". But we can obviously save time on not calling API at all if the header is not present.

bridge/engine/ActiveMediaList.h
reddvl1980: Shall we remove all `_` prefixes here? it's a struct, members are not private.

bridge/engine/AudioForwarderReceiveJob.cpp
reddvl1980: Fixed.

bridge/engine/ActiveMediaList.h
reddvl1980: I think we need some flag to indicate how to proceed with regards of active talker detection. At any point of time, if the request comes to get a set of active talkers, we'll check the flag, and either use the limited set of endpoints that have `isPtt` flag, _or_ we use our own ranking.

bridge/engine/AudioForwarderReceiveJob.cpp
reddvl1980: Fixed.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -355,12 +384,35 @@ void ActiveMediaList::process(const uint64_t timestampMs, bool& outDominantSpeak
...
 
+    if (!_c9_conference)
+    {
+        _activeTalkers.clear();
+    }
+
     for (size_t i = 0; i < _audioLastN && !heap.empty(); ++i)
     {
-        updateActiveAudioList(heap.top().participant);
+        auto top = heap.top();
```
olofkallander, line 394: use auto& to avoid having a copy here. Not sure optimizer will fix that. I have seen "mistakes" I have made where it is by value if not explicitly specified.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -355,12 +384,35 @@ void ActiveMediaList::process(const uint64_t timestampMs, bool& outDominantSpeak
...
+        updateActiveAudioList(top.participant);
+        if (!_c9_conference)
+        {
+            if (top.score > top.noisePlus20percent)
```
olofkallander, line 398: I would rather calculate the noise floor * 1.2 here to avoid name that includes "calculation". Just call it noiseLevel and let rankSpeakers do only what the name implies, rank Speakers. If you change the margin later, we do not want to change the name.

olofkallander: A question regarding the margin. We have a score based on diff in dB basically. Is it the right thing to *1.2 or is a +12 more real?

bridge/engine/ActiveMediaList.cpp
```diff
@@ -355,12 +384,35 @@ void ActiveMediaList::process(const uint64_t timestampMs, bool& outDominantSpeak
...
+    {
+        _activeTalkers.clear();
+    }
+
     for (size_t i = 0; i < _audioLastN && !heap.empty(); ++i)
     {
-        updateActiveAudioList(heap.top().participant);
+        auto top = heap.top();
+        updateActiveAudioList(top.participant);
+        if (!_c9_conference)
```
olofkallander, line 396: An interesting angle on this. What if we have a hybrid conference, in the name of integration. Some users produce the ptt and some do not.
Any user that is not producing c9 hdr would be excluded from the active talker list.

bridge/Mixer.cpp
```diff
@@ -768,32 +768,45 @@ std::unordered_set<std::string> Mixer::getEndpoints() const
...
-                (audio->second) && audio->second->endpointIdHash == _engineMixer.getDominantSpeakerId();
+            foundAudio = true;
+            endpoint.isDominantSpeaker = audio->second->endpointIdHash == _engineMixer.getDominantSpeakerId();
+            if (activeTalkers.find(audio->second->endpointIdHash) != activeTalkers.end())
+            {
+                endpoint.isActiveTalker = true;
+            }
+            else
+            {
+                endpoint.isActiveTalker = endpoint.isDominantSpeaker;
```
olofkallander, line 798: shouldn't the dominant speaker also be in the active talker list.

bridge/engine/ActiveMediaList.cpp
reddvl1980: fixed! good catch

bridge/engine/ActiveMediaList.cpp
reddvl1980: We've discussed it with Shanling, and come to conclusion that mixed conference will never happen. It's in the doc I shared.

bridge/engine/ActiveMediaList.cpp
reddvl1980: Fixed to configurable dbDiff value (default is tested 18 dB)

bridge/Mixer.cpp
reddvl1980: In the light of what we saw in the testing, I thing we need to keep them separate. Situation might be where there is a dominant speaker (added by default) but no active speakers.

config/Config.h
```diff
@@ -29,6 +29,7 @@ class Config : public ConfigReader
     CFG_PROP(int32_t, silenceThresholdLevel, 127);
     CFG_PROP(uint32_t, lastN, 3);
     CFG_PROP(uint32_t, lastNextra, 2);
+    CFG_PROP(uint32_t, activeTalkerDbDiff, 18);
```
olofkallander, line 32: Maybe DbDiff is not saying much. Diff to what?
activeTalkerSilenceThresholdDb

test/integration/IntegrationTest.cpp
```diff
@@ -234,19 +234,17 @@ TEST_F(IntegrationTest, plain)
     EXPECT_TRUE(endpointRequest.isSuccess());
     EXPECT_TRUE(endpointRequest.getJsonBody().is_array());
 
+    logger::debug("!!! %s", "###", endpointRequest.getResponse().c_str());
```
olofkallander, line 237: should remove this right?

bridge/Mixer.cpp
```diff
@@ -768,32 +768,45 @@ std::unordered_set<std::string> Mixer::getEndpoints() const
...
             auto transport = audio->second->transport;
             endpoint.iceState = transport->getIceState();
             endpoint.dtlsState = transport->getDtlsState();
         }
     }
 
-    endpoint.hasVideo = _videoStreams.find(endpointId) != _videoStreams.cend() ? true : false;
-    endpoint.isBundled = _bundleTransports.find(endpointId) != _bundleTransports.cend() ? true : false;
-    endpoint.isRecording = _recordingStreams.find(endpointId) != _recordingStreams.cend() ? true : false;
+    bool foundVideo = _videoStreams.find(endpointId) != _videoStreams.cend() ? true : false;
```
olofkallander, line 807: No need for ?
The expression is already true or false.

bridge/Mixer.cpp
```diff
@@ -768,32 +768,45 @@ std::unordered_set<std::string> Mixer::getEndpoints() const
...
     if (audio != _audioStreams.cend())
     {
-        endpoint.hasAudio = true;
         if (audio->second)
         {
-            endpoint.isActiveSpeaker =
-                (audio->second) && audio->second->endpointIdHash == _engineMixer.getDominantSpeakerId();
+            foundAudio = true;
+            endpoint.isDominantSpeaker = audio->second->endpointIdHash == _engineMixer.getDominantSpeakerId();
+            if (activeTalkers.find(audio->second->endpointIdHash) != activeTalkers.end())
```
olofkallander, line 792: Just assign endpoint.isActiveTalker to the result of comparison.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -280,13 +284,25 @@ void ActiveMediaList::updateLevels(const uint64_t timestampMs)
...
-size_t ActiveMediaList::rankSpeakers(float& currentDominantSpeakerScore)
+size_t ActiveMediaList::rankSpeakers(float& currentDominantSpeakerScore,
+    TActiveTalkersSnapshot& outActiveTalkersSnapshot)
 {
     currentDominantSpeakerScore = 0;
 
     size_t speakerCount = 0;
     for (auto& participantLevelEntry : _audioParticipants)
     {
+        if (_c9_conference)
```
olofkallander, line 295: Perhaps && is better than nesting in this case.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -355,11 +387,25 @@ void ActiveMediaList::process(const uint64_t timestampMs, bool& outDominantSpeak
...
 
+    assert(_c9_conference || 0 == activeTalkersSnapshot.count);
+
     for (size_t i = 0; i < _audioLastN && !heap.empty(); ++i)
     {
-        updateActiveAudioList(heap.top().participant);
+        const auto& top = heap.top();
+        updateActiveAudioList(top.participant);
+        if (!_c9_conference)
+        {
```
olofkallander, line 397: Perhaps && instead of nesting

config/Config.h
reddvl1980: Changed.

test/integration/IntegrationTest.cpp
reddvl1980: yep ;)

bridge/Mixer.cpp
reddvl1980: Done.

bridge/Mixer.cpp
reddvl1980: Fixed and inlined.

bridge/engine/ActiveMediaList.cpp
reddvl1980: Isn't it a bit strange that I understood what you wrote actually :) Fixed.

olofkallander (approved):

bridge/engine/ActiveMediaList.cpp
reddvl1980: agree

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -132,23 +132,45 @@ void AudioForwarderReceiveJob::run()
...
+        bool silence = false;
+        if (isPtt.isSet())
+        {
+            silence = !isPtt.get();
+            _activeMediaList.onNewPtt(_sender->getEndpointIdHash(), isPtt.get());
+            _activeMediaList.onNewAudioLevel(_sender->getEndpointIdHash(), silence ? 127 : 0);
+        }
+        else
+        {
+            if (audioLevel.isSet())
```
olofkallander, line 162: should be 
else if(audioLevel.isSet()) 
But now ptt has precedence over audio level since ptt.isSet is chcked first.

olofkallander: If we have audio level we should report it regardless of ptt.
If we have ptt but not audiolevel, we can fake audiolevel as you do

olofkallander: My guess, is that if we have both, ptt is going to be 1 along with fairly high audio level. 
If ptt is 1 and audio level low, I prefer if someone else that has audio is heard.

test/integration/IntegrationTest.cpp
```diff
@@ -31,6 +31,27 @@
 #include <sstream>
 #include <unordered_set>
 
+#define DEFINE_3_CLIENT_CONFERENCE(TChannel, BASE_URL)                                                                 \
```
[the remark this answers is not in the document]
reddvl1980: yep, integration tests has "a lot of potential" for refactoring.

bridge/engine/AudioForwarderReceiveJob.cpp
reddvl1980: Ok, reversing precendence, now we have:
```
        if (audioLevel.isSet())
        {
            silence = audioLevel.get() > _silenceThresholdLevel;
        }
        else if (isPtt.isSet())
        {
            silence = !isPtt.get();
            audioLevel.set(silence ? 127 : 0);
            _activeMediaList.onNewPtt(_sender->getEndpointIdHash(), isPtt.get());
        }
```

PR #115, opened 2022-07-25
Olofk/pktmeta

olofkallander:


bridge/engine/ActiveMediaList.cpp
```diff
@@ -217,63 +220,64 @@ bool ActiveMediaList::removeVideoParticipant(const size_t endpointIdHash)
...
     {
-        auto participantLevelItr = _audioParticipants.find(levelEntry._participant);
+        auto participantLevelItr = _audioParticipants.find(levelEntry.participant);
         if (participantLevelItr == _audioParticipants.end())
         {
             continue;
         }
 
         auto& participantLevels = participantLevelItr->second;
+        participantLevels.lastPacketReceived = timestamp;
```
olofkallander, line 249: This is reason I wanted to change to abs Timestamp. The Ms is a slight risk too as it will wrap differently from absTimestamp. In case the absTimeStamp starts at unfortunate value.

bridge/MixerManager.cpp
```diff
@@ -82,18 +82,26 @@ Mixer* MixerManager::create(uint32_t lastN)
...
     {
         audioSsrcs.push_back(_ssrcGenerator.next());
     }
 
-    // Last-n + screen share, extra
-    for (uint32_t i = 0; i < lastN + 3; ++i)
+    // screen share / slides
     {
-        videoSsrcs.push_back({_ssrcGenerator.next(), _ssrcGenerator.next()});
+        SimulcastLevel a[1] = {{_ssrcGenerator.next(), _ssrcGenerator.next(), false}};
```
olofkallander, line 95: Generate 6 target ssrcs per rewrite target. Before we had only 2 ssrcs as each client only receives one vieo stream from each forwarded participant. That is still true, but we will need to be ready to forward 3 streams per forwarded participant over the barbell.

bridge/Mixer.cpp
```diff
@@ -839,8 +837,8 @@ bool Mixer::getVideoStreamDescription(const std::string& endpointId, StreamDescr
     {
         for (auto ssrcPair : _videoSsrcs)
         {
-            outDescription._localSsrcs.push_back(ssrcPair._ssrc);
-            outDescription._localSsrcs.push_back(ssrcPair._feedbackSsrc);
+            outDescription._localSsrcs.push_back(ssrcPair.levels[0]._ssrc);
```
olofkallander, line 840: We still only show first ssrc to normal clients.

bridge/Mixer.cpp
```diff
@@ -2297,6 +2298,23 @@ bool Mixer::configureBarbellTransport(const std::string& barbellId,
...
+{
+    std::lock_guard<std::mutex> locker(_configurationLock);
+    auto barbellItr = _barbells.find(barbellId);
+    if (barbellItr == _barbells.end())
+    {
+        return false;
+    }
+
+    auto& barbell = barbellItr->second;
+    barbell->audioSsrcs = audioSsrcs;
```
olofkallander, line 2313: The barbell will know about all video and audio ssrcs coming in over the barbell.

bridge/engine/EngineBarbell.h
```diff
@@ -33,6 +53,31 @@ struct EngineBarbell
...
+        utils::Optional<size_t> endpointIdHash;
+        utils::Optional<std::string> endpointId;
+    };
+
+    // inbound ssrcs over barbell and how they group together and map currently to remote endpointId
+    std::vector<SsrcGroup> videoStreams;
+    std::vector<AudioStream> audioStreams;
+    SsrcGroup slideStream;
+
+    concurrency::MpmcHashmap32<uint32_t, SsrcGroup*> videoSsrcMap;
```
olofkallander, line 79: will contain all 6 ssrcs for a video stream as items pointing to the SsrcGroup.

bridge/engine/EngineBarbell.h
```diff
@@ -33,6 +53,31 @@ struct EngineBarbell
...
     // some map for ssrc to user id endpointIdHash to be used in activemediaList, that we update from data channel
     // messages
+
+    struct SsrcGroup
+    {
+        uint32_t ssrcs[3];
+        uint32_t feedbackSsrcs[3];
+        uint32_t count = 3;
+
+        utils::Optional<size_t> endpointIdHash;
```
olofkallander, line 63: This is where we will store the actual endpoint mapping to the video channel.

bridge/engine/EngineBarbell.h
```diff
@@ -33,6 +53,31 @@ struct EngineBarbell
...
+        utils::Optional<std::string> endpointId;
```
olofkallander, line 64: This string is disputable as we have potential mem alloc here in RT world.
When UMM is received over babrell data channel we will want to update this string and the idhash. It will be used later to create the UMM for clients.

bridge/engine/ActiveMediaList.cpp
olofkallander: This is reason I wanted to change to abs Timestamp. The Ms is a slight risk too as it will wrap differently from absTimestamp. In case the absTimeStamp starts at unfortunate value.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -646,7 +647,7 @@ bool ActiveMediaList::makeUserMediaMapMessage(const size_t lastN,
         const auto rewriteMapItr = _videoSsrcRewriteMap.find(videoListEntry->_data);
```
olofkallander, line 647: This is incorrect. The UMM for clients must also check the video stream endpoints in the barbells.
Unless we add fake video streams there due to UMM over the barbell.

api/JsonBuilder.h
```diff
@@ -0,0 +1,133 @@
...
+        {
+            _builder.append(",");
+        }
+
+        _builder.append("{");
+    };
+
+    ~JsonObject() { _builder.append("}"); }
+
+    void addProperty(const char* name, const char* value)
```
reddvl1980, line 61: Consider using generic "Builder" pattern, where method adding properties or sub-builders returns *this (as JsonObjectBuilder) and only final "build()" returns  final constructed object. That would allow to chain things, e.g.:

JsonObjectBuilder builder;
JsonObject obj = builder.addProperty("foo", "bar")
                        .addProperty(...)
                        ...
                        .addArray(...)
                              .addElem(...)
                              ...
                              .build()
                        .build();

bridge/BarbellStreamGroupDescription.h
```diff
@@ -0,0 +1,27 @@
...
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace bridge
+{
+
+struct BarbellStreamGroupDescription
+{
+    BarbellStreamGroupDescription() : slides(false){};
```
reddvl1980, line 12: Comment is needed here regarding "slides" ambiguity.

bridge/endpointActions/BarbellActions.cpp
```diff
@@ -170,8 +185,35 @@ httpd::Response configureBarbell(ActionContext* context,
             utils::format("Failed to configure barbell transport %s - %s", conferenceId.c_str(), barbellId.c_str()));
     }
 
-    mixer->addBarbellToEngine(barbellId);
+    std::vector<BarbellStreamGroupDescription> videoDescriptions;
+    auto& videoDescription = barbellDescription._video.get();
+    for (size_t i = 0; i < videoDescription._ssrcGroups.size(); i += 2)
```
reddvl1980, line 190: The fact that we're assuming correct interleaving of SSRCs is bothering me, it does not immediately screams "not robust" but it is close to. Why not to use dedicated arrays (e.g.: mediaSsrc, feedbackSsrc) and getters like you did for BarbellStreamGroupDescription?

reddvl1980:
LGTM, but let's wait 1-2 days for my C9 change merged.

api/JsonBuilder.h
olofkallander: Considered. The purpose of these objects are to leverage the constructor destructor to manage the braces and create a structure in code that represents the json.
lohmann json would have been better but it entails mem alloc and exceptions.

bridge/endpointActions/BarbellActions.cpp
olofkallander: Totally agree. It hurts to leave it like this, but I created a ticket to refactor this specifically. It would create a lot of changes in EndpointDescripton etc to facilitate code reuse.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -664,6 +664,81 @@ bool ActiveMediaList::makeUserMediaMapMessage(const size_t lastN,
     return true;
 }
 
+bool ActiveMediaList::makeBarbellUserMediaMapMessage(
+    const concurrency::MpmcHashmap32<size_t, EngineAudioStream*>& engineAudioStreams,
+    const concurrency::MpmcHashmap32<size_t, EngineVideoStream*>& engineVideoStreams,
+    utils::StringBuilder<1024>& outMessage)
+{
+    JsonObject umm(outMessage);
+    umm.addProperty("type", "user-media-map");
```
RicardoMDomingues, line 673: I understand the utility but feels strange have a destructor that will add the final bracket and the data stay on the variable passed on constructor.

Also `JsonObject` is confusing. I guess we should consider a better name for it. Unfortunately I am terrible to find goof names :) but I have some ideas:`JsonObjectEncoder`, `JsonObjectWriter`...

olofkallander:
> what is the difference in meaning of an empty string and an optional string? 

You are right. In later PR it has changed type to an in place fixed string. I will remove the Optional and use empty() instead to check if it is not set.

olofkallander:
> Does not sound a good assumption. Example, currently there is an issue on chrome 103 where we lost the simulcast, we can't be sure that a single video ssrc means it is the `_videoScreenShareSsrc` 

These are the rewrite ssrcs that we create within SMB. We know that the one with single ssrc is intended for screen share. The others always have 3 each to allow for barbell send. For normal client the first ssrc is used, but it does not matter how many ssrcs the client presents. I think mobile presents 2 for example.

olofkallander:
> I understand the utility but feels weird having a destructor that will add the final bracket and the data stay on the variable passed on constructor.

Also `JsonObject` is confusing. I guess we should consider a better name for it. Unfortunately I am terrible to find good names :) but I have some ideas:`JsonObjectEncoder`, `JsonObjectWriter`... and change the methods to `encodeProperty`/`writeProperty`

Also would be nice to have a free function to make the type deduction for us. Like:
```
template<class TStringBuilder>
JsonObjectWriter<TStringBuilder> startWriteJsonObject(TStringBuilder& builder)
{
    return JsonObjectWriter<TStringBuilder>(builder);
}
```

and use it like this:
```
auto jsonObjectWriter = startWriteJsonObject(outMessage);
```

And maybe a method that would return new json writers to use like this:

```
auto videoArrayWriter = jsonObjectWriter.startWriteArrayProperty("video-endpoints");
{
     auto videoEp = videoArrayWriter.startWriteObject();
     videoEp.writeProperty("endpoint-id", ...)

}
```

 If the namespace now changed to json::builder, I think even Object and Array are good names.
json::builder::Object
json::builder::Array

RicardoMDomingues:
> > I understand the utility but feels weird having a destructor that will add the final bracket and the data stay on the variable passed on constructor.
> 
> Also `JsonObject` is confusing. I guess we should consider a better name for it. Unfortunately I am terrible to find good names :) but I have some ideas:`JsonObjectEncoder`, `JsonObjectWriter`... and change the methods to `encodeProperty`/`writeProperty`
> 
> Also would be nice to have a free function to make the type deduction for us. Like:
> 
> ```
> template<class TStringBuilder>
> JsonObjectWriter<TStringBuilder> startWriteJsonObject(TStringBuilder& builder)
> {
>     return JsonObjectWriter<TStringBuilder>(builder);
> }
> ```
> 
> and use it like this:
> 
> ```
> auto jsonObjectWriter = startWriteJsonObject(outMessage);
> ```
> 
> And maybe a method that would return new json writers to use like this:
> 
> ```
> auto videoArrayWriter = jsonObjectWriter.startWriteArrayProperty("video-endpoints");
> {
>      auto videoEp = videoArrayWriter.startWriteObject();
>      videoEp.writeProperty("endpoint-id", ...)
> 
> }
> ```
> 
> If the namespace now changed to json::builder, I think even Object and Array are good names. json::builder::Object json::builder::Array

Yeah, using it always with the namespace it can work. My only minor concern is about the word "builder". "builder" is being used on programming to have the builder pattern as @reddvl1980 pointed, as the goal of this is not being a builder in that sense but more an encoded writer, perhaps the word "builder" is not the best

PR #117, opened 2022-07-26
Implement "Active Talkers" detection.

reddvl1980:


bridge/engine/ActiveMediaList.h
```diff
@@ -160,6 +168,7 @@ class ActiveMediaList
         int32_t _nonZeroLevelsShortWindow;
         float _maxRecentLevel;
         float _noiseLevel;
+        bool _isPtt;
```
olofkallander, line 171: should we call this pushToTalk instead?
To avoid abbreviations that are confusing to .. anyone

bridge/Mixer.cpp
```diff
@@ -768,32 +768,48 @@ std::unordered_set<std::string> Mixer::getEndpoints() const
...
+bool Mixer::getEndpointExtendedInfo(const std::string& endpointId,
+    api::ConferenceEndpointExtendedInfo& endpoint,
+    const std::unordered_set<size_t>& activeTalkers)
+{
+    if (!getEndpointInfo(endpointId, endpoint, activeTalkers))
+        return false;
+
+    const auto audio = _audioStreams.find(endpointId);
+    const auto transport = audio->second->transport;
+    // transport->getRemotePeer().
```
olofkallander, line 811: Is he ok with not getting this info now?
I guess he was interested in remote ip

PR #118, opened 2022-07-27
Implement simple exception-free JSON Parser

reddvl1980:
- it is assumed that JSON is valid (simplified checks exist, but do not cover everything)
- values are addressed via dot-delimited path (e.g. "transport.ice.candidates")
- arrays are retrieved in std::vector<SimpleJson>
- per-object cache is used for successful partial path searches

### DIFF
--- a/utils/SimpleJson.cpp
+++ b/utils/SimpleJson.cpp
@@ -67,6 +67,7 @@ SimpleJson::Type SimpleJson::acquirePrimitiveType()
         return Type::Boolean;
     }
     cursor = findNumberEnd(_cursorIn);
+    char _buffer[33];
     if (cursor == _cursorOut && size() < sizeof(_buffer))
     {
         strncpy(_buffer, _cursorIn, size());
@@ -323,23 +324,25 @@ SimpleJson SimpleJson::find(const std::string& path) const
         return (property.getType() == SimpleJson::Type::Object) ? property.find(token.next) : SimpleJsonNone;
 }
 
-bool SimpleJson::getValue(int64_t& out)
+bool SimpleJson::getValue(int64_t& out) const
 {
     if (Type::Integer != _type)
         return false;
+    char _buffer[33];
     strncpy(_buffer, _cursorIn, size());
     return 1 == sscanf(_buffer, "%" SCNd64, &out);
 }
 
-bool SimpleJson::getValue(double& out)
+bool SimpleJson::getValue(double& out) const
 {
     if (Type::Float != _type)
         return false;
+    char _buffer[33];
     strncpy(_buffer, _cursorIn, size());
     return 1 == sscanf(_buffer, "%lf", &out);
 }
 
-bool SimpleJson::getValue(std::string& out)
+bool SimpleJson::getValue(std::string& out) const
 {
     if (Type::String != _type || size() < 2)
         return false;
@@ -347,7 +350,7 @@ bool SimpleJson::getValue(std::string& out)
     return true;
 }
 
-bool SimpleJson::getValue(bool& out)
+bool SimpleJson::getValue(bool& out) const
 {
     if (Type::Boolean != _type)
         return false;
@@ -364,7 +367,7 @@ bool SimpleJson::getValue(bool& out)
     return false;
 }
 
-bool SimpleJson::getValue(std::vector<SimpleJson>& out)
+bool SimpleJson::getValue(std::vector<SimpleJson>& out) const
 {
     if (Type::Array != _type || '[' != *_cursorIn)
         return false;
--- a/utils/SimpleJson.cpp
+++ b/utils/SimpleJson.cpp
@@ -3,8 +3,6 @@
 #include <inttypes.h>
 #include <math.h>
 
-#define SIMPLE_JSON_CACHE_DIAG 0
-
 namespace utils
 {
 template <size_t SIZE>
--- a/utils/SimpleJson.cpp
+++ b/utils/SimpleJson.cpp
@@ -1,5 +1,7 @@
 #include "SimpleJson.h"
 #include "StringTokenizer.h"
+#include <inttypes.h>
+#include <math.h>
 
 namespace utils
 {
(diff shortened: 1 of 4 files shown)


utils/SimpleJson.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+namespace utils
+{
+
+const SimpleJson SimpleJson::SimpleJsonNone = SimpleJson::createJsonNone();
+
+SimpleJson SimpleJson::create(const char* cursorIn, size_t length)
+{
+    if (!cursorIn || 0 == length)
+        return SimpleJsonNone;
```
olofkallander, line 14: Add {} to all of these

utils/SimpleJson.cpp
```diff
@@ -0,0 +1,410 @@
...
+    }
+    return end;
+}
+
+const char* SimpleJson::findMatchEnd(const char* start, const std::string& match) const
+{
+    if (start + match.length() - 1 > _cursorOut)
+        return nullptr;
+    return strncmp(start, match.c_str(), match.length()) ? nullptr : start + match.length() - 1;
+}
```
olofkallander, line 157: more space between methods

utils/SimpleJson.h
```diff
@@ -0,0 +1,104 @@
...
+#include <string>
+#include <vector>
+
+namespace utils
+{
+
+class SimpleJson
+{
+public:
+    enum Type
```
olofkallander, line 14: The type is strictly not necessary. The type is mostly obvious from the start character of the SimpleJson item.
The only ambiguity that would require more processing of characters is when it is a number. We cannot tell if it is a float or integer until we have looked for decimal point. Even then it is up to the user to retrieve it as a double.

utils/SimpleJson.h
```diff
@@ -0,0 +1,104 @@
...
+    }
+    SimpleJson& operator=(const SimpleJson& ref)
+    {
+        _cursorIn = ref._cursorIn;
+        _cursorOut = ref._cursorOut;
+        _type = ref._type;
+        return *this;
+    }
+
+    Type getType() const { return _type; }
```
olofkallander, line 43: I wonder if this variant like typing is benefitial. 
One thing I lack here is typed JsonArray that has begin end methods so I can iterate over the array items.
Putting that into a generic SimpleJson seems a bit akwars as the object is usually not an array.

utils/SimpleJson.cpp
```diff
@@ -0,0 +1,410 @@
...
+    auto cursor = start;
+    if (!start || OPEN_CHAR != *start)
+        return end;
+
+    int level = 1;
+    while (++cursor < _cursorOut)
+    {
+        switch (*cursor)
+        {
+        case OPEN_CHAR:
```
olofkallander, line 133: This is dangerous parsing of subobjects. If there is a string with a } in, we are toast.

utils/SimpleJson.cpp
```diff
@@ -0,0 +1,410 @@
...
+{
+    auto token = StringTokenizer::tokenize(path.c_str(), path.length(), '.');
+    if (token.empty() || Type::Object != _type)
+        return SimpleJsonNone;
+
+    if (!cachedPath.empty())
+    {
+        cachedPath += '.';
+    }
+    cachedPath += token.str();
```
olofkallander, line 329: this is building with new allocations

utils/SimpleJson.cpp
olofkallander: This is dangerous parsing of subobjects. If there is a string with a } in, we are toast.

utils/SimpleJson.h
```diff
@@ -0,0 +1,104 @@
...
+
+    void validateFast();
+    Type acquirePrimitiveType();
+    size_t size() const { return _cursorOut - _cursorIn + 1; }
+
+private:
+    const char* _cursorIn;
+    const char* _cursorOut;
+    Type _type;
+    std::map<std::string, SimpleJson> _nodeCache;
```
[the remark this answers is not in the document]
olofkallander: this std:map will also allocate memory at will.

utils/SimpleJson.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+    auto propIt = nodeCache.find(cachedPath);
+    auto property = propIt == nodeCache.end() ? findProperty(_cursorIn + 1, token.str()) : propIt->second;
```
olofkallander, line 332: These are typically difficult to read. Help the eye with braces
auto property = (propIt == nodeCache.end() ? findProperty(_cursorIn + 1, token.str()) : propIt->second);

utils/SimpleJson.cpp
```diff
@@ -0,0 +1,410 @@
...
+    auto node = _nodeCache.find(path);
+    if (node != _nodeCache.end())
+    {
+        return node->second;
+    }
+    std::string cachedPath = "";
+    return findInternal(path, cachedPath, _nodeCache);
+}
+
+SimpleJson SimpleJson::findInternal(const std::string& path,
```
olofkallander, line 317: Insted of this advanced token spliting find I would have starte with a simpler findProperty, that first looks through the char* in index[] to see if any of those matches with "property-name". Then eat recursively through the last property value to find the next property and add it to the index[]. Eventually I find it or hit }

utils/SimpleJson.cpp
```diff
@@ -0,0 +1,410 @@
...
+    }
+    if (size() == 5 && !strncmp(_cursorIn, "false", 5))
+    {
+        out = false;
+        return true;
+    }
+    return false;
+}
+
+bool SimpleJson::getValue(std::vector<SimpleJson>& out) const
```
olofkallander, line 385: This is also memory alloc. JsonArray with its own iterator, is not mem alloc.

utils/SimpleJson.cpp
reddvl1980: Can't put braces where put, but possible here:
`auto property = (propIt == nodeCache.end()) ? findProperty(_cursorIn + 1, token.str()) : propIt->second;`

utils/SimpleJson.h
reddvl1980: Removed. Changed to per-type getters returnong Optional<>

utils/SimpleJson.h
reddvl1980: They type is derived automatically while linear scanning, and helps to whether getter is eligible. Also, it helps to make a special SimpleJsonNone value.

utils/SimpleJson.cpp
reddvl1980: Done.

utils/SimpleJson.cpp
reddvl1980: Method literally does finds end of the Object {}  or Array [] while "fast forwarding" all inside. We are not parsing anything, but we have to be careful about strings, thus if we detect '"' we skip until the end of the string taking into account escaping.

utils/SimpleJson.cpp
reddvl1980: Fixed.

utils/SimpleJson.cpp
reddvl1980: Fixed. All std::string operations removed. CachedPath is constructed in a fixed size buffer, which should be enough for our applications.

utils/SimpleJson.h
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    Type getType() const { return _type; }
+    SimpleJson find(const char* const path);
+
+    Optional<int64_t> getIntValue() const;
+    Optional<double> getFloatValue() const;
+    bool getStringValue(const char*& out, size_t& outLen) const;
+    Optional<bool> getBoolValue() const;
+    Optional<SimpleJsonArray> getArrayValue() const;
```
olofkallander, line 100: getArray()

utils/SimpleJson.cpp
```diff
@@ -0,0 +1,522 @@
...
+        return Optional<bool>(true);
+    }
+    if (size() == 5 && !strncmp(_cursorIn, "false", 5))
+    {
+        return Optional<bool>(false);
+    }
+    return Optional<bool>();
+}
+
+Optional<SimpleJsonArray> SimpleJson::getArrayValue() const
```
olofkallander, line 494: I think we should do lazy evaluation. Array does not need internal data storage. The iterators can just find the next lement position when we do ++.

PR #119, opened 2022-07-29
RTC-13096 Extended Endpoint Info API

reddvl1980:


api/Parser.cpp
```diff
@@ -517,6 +517,32 @@ std::vector<ConferenceEndpoint> parseConferenceEndpoints(const nlohmann::json& r
     return endpoints;
 }
 
+ConferenceEndpointExtendedInfo parseEndpointExtendedInfo(const nlohmann::json& data)
+{
+    ConferenceEndpointExtendedInfo endpoint;
+    endpoint.basicEndpointInfo = parseConferenceEndpoint(data);
+    nlohmann::json iceSelectedTuple = nlohmann::json::object();
+
+    setIfExistsOrThrow<>(iceSelectedTuple, data, "iceSelectedTuple");
```
olofkallander, line 526: Isn't it selected "candidate"?

bridge/engine/ActiveMediaList.h
```diff
@@ -63,13 +63,16 @@ class ActiveMediaList
...
     void onNewPtt(const size_t endpointIdHash, bool isPtt);
+    void mapSsrc2Usid(uint32_t ssrc, uint32_t usid);
 
     void process(const uint64_t timestampMs, bool& outDominantSpeakerChanged, bool& outUserMediaMapChanged);
 
     inline size_t getDominantSpeaker() const { return _dominantSpeakerId; }
 
     const std::unordered_set<size_t> getActiveTalkers() const;
 
+    utils::Optional<uint32_t> getUsid(const uint32_t ssrc);
```
olofkallander, line 74: I do not understand the abbreviation. Can we spell it out?

bridge/engine/ActiveMediaList.h
```diff
@@ -219,6 +222,7 @@ class ActiveMediaList
     concurrency::MpmcQueue<AudioLevelEntry> _incomingAudioLevels;
     concurrency::MpmcQueue<uint32_t> _audioSsrcs;
     concurrency::MpmcHashmap32<size_t, uint32_t> _audioSsrcRewriteMap;
+    concurrency::MpmcHashmap32<uint32_t, uint32_t> _audioSsrcUsidMap;
```
olofkallander, line 225: Why do we store it in ActiveMediaList? It is not really related to active streams.
Active talkers, dominant speaker, and map speaker to outbound ssrc is here. But the rest should be in EngineMixer or Mixer.

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -137,12 +137,18 @@ void AudioForwarderReceiveJob::run()
 
         utils::Optional<uint8_t> audioLevel;
         utils::Optional<bool> isPtt;
+        uint32_t usid = 0;
```
olofkallander, line 140: userId, or even c9UserId.

transport/TransportImpl.h
```diff
@@ -172,6 +172,7 @@ class TransportImpl : public RtcTransport,
     void runTick(uint64_t timestamp) override;
     ice::IceSession::State getIceState() const override { return _iceState; };
     SrtpClient::State getDtlsState() const override { return _dtlsState; };
+    utils::Optional<ice::TransportType> getSelectedTransportType() const override;
```
olofkallander, line 175: I wonder if we should consider the option of having this in the DataReceiver callback interface. 
It feels like Transport is getting fatter and fatter with member variables and getters. 
We should perhaps give it a thought to evaluate the consequences. Perhaps this is the most light weight solution but we have more and more access methods to retrieve a little piece of the state. One option is to read out a struct over state using MpmcPublish, which then contains more, in synch, state information and is cheaper to read once from atomic than doing multiple atomic reads.

transport/TransportImpl.cpp
```diff
@@ -2047,6 +2055,9 @@ void TransportImpl::onIceStateChanged(ice::IceSession* session, const ice::IceSe
                 _peerRtpPort = candidatePair.second.address;
                 _peerRtcpPort = candidatePair.second.address;
 
+                _transportType = endpoint->getTransportType();
+                _transportTypeIsSet = true;
```
olofkallander, line 2059: Shouldn't the _transportType be an Optional<> in the first place?

api/Parser.cpp
```diff
@@ -517,6 +517,32 @@ std::vector<ConferenceEndpoint> parseConferenceEndpoints(const nlohmann::json& r
...
+    setIfExistsOrThrow<>(endpoint.localIP, iceSelectedTuple, "localIP");
+    setIfExistsOrThrow<>(endpoint.remoteIP, iceSelectedTuple, "remoteIP");
+    setIfExistsOrThrow<>(endpoint.localPort, iceSelectedTuple, "localPort");
+    setIfExistsOrThrow<>(endpoint.remotePort, iceSelectedTuple, "remotePort");
+    setIfExistsOrThrow<>(endpoint.protocol, iceSelectedTuple, "protocol");
+
+    for (const auto& it : requiredJsonArray(data, "audioUsidToSsrcMap"))
```
olofkallander, line 533: audioUserIdToSsrcMap ?

api/Parser.cpp
reddvl1980: No, it's a so-called 5-tuple {ip,port,ip,port,protocol} selected by ICE. Network-related essential info from both selected local and remote candidates.

transport/TransportImpl.cpp
reddvl1980: Fixed.

transport/TransportImpl.h
reddvl1980: I'd rather have it in a separate MR. What do you think?

api/Parser.cpp
```diff
@@ -517,6 +517,32 @@ std::vector<ConferenceEndpoint> parseConferenceEndpoints(const nlohmann::json& r
...
+    {
+        for (const auto& inner : it.items())
+        {
+            auto usid = inner.key();
```
RicardoMDomingues, line 538: [minor] Avoid a `std:string` copy
```suggestion
            const auto& usid = inner.key();
```

bridge/engine/AudioForwarderReceiveJob.cpp
reddvl1980: Fixed.

bridge/engine/ActiveMediaList.h
reddvl1980: I renamed it ot getUserId() because I am hesitatant to put 'c9' into the name. But that would be an alternative: getC9userId().

RicardoMDomingues (approved):
Apart from Olof comments, LGTM. Only added some minor improvement comments

bridge/Mixer.cpp
```diff
@@ -788,6 +790,10 @@ bool Mixer::getEndpointInfo(const std::string& endpointId,
         if (audio->second)
         {
             foundAudio = true;
+            if (audio->second->remoteSsrc.isSet())
+            {
+                ssrc.set(audio->second->remoteSsrc.get());
+            }
```
RicardoMDomingues, line 796: [Minor] If `audio->second->remoteSsrc.isSet()`, `ssrc` should be empty also, then let's simplify the code:
```suggestion
            ssrc = audio->second->remoteSsrc;
```

api/Parser.cpp
reddvl1980: Fixed.

bridge/Mixer.cpp
reddvl1980: Agree

olofkallander (approved):
Yes, perhaps we should create a ticket for general slim down of Transport with regards to to information gathered. I know stats is also a thing that has been refactored several times to make it safer and simpler to retrieve.

api/Parser.cpp
```diff
@@ -517,6 +517,36 @@ std::vector<ConferenceEndpoint> parseConferenceEndpoints(const nlohmann::json& r
     return endpoints;
 }
 
+ConferenceEndpointExtendedInfo parseEndpointExtendedInfo(const nlohmann::json& data)
+{
+    ConferenceEndpointExtendedInfo endpoint;
+    endpoint.basicEndpointInfo = parseConferenceEndpoint(data);
+    nlohmann::json iceSelectedTuple = nlohmann::json::object();
+    logger::debug("!!! %s", "###", data.dump(4).c_str());
```
olofkallander, line 525: should be removed

bridge/engine/ActiveMediaList.cpp
```diff
@@ -391,7 +391,7 @@ void ActiveMediaList::process(const uint64_t timestampMs, bool& outDominantSpeak
     {
         const auto& top = heap.top();
         updateActiveAudioList(top.participant);
-        if (_c9_conference && top.score - top.noiseLevel > _activeTalkerSilenceThresholdDb)
+        if (!_c9_conference && top.score - top.noiseLevel > _activeTalkerSilenceThresholdDb)
```
olofkallander, line 394: Are we not reporting a fake audio level when ptt=1? 
Then why do we build activeTalkersSnapshot in two places. This condition should suffice for both c9 and !c9. 
Instead we remove the adding to activeTalkersSnapshot in rankSpeakers.

bridge/engine/ActiveMediaList.h
```diff
@@ -70,6 +70,8 @@ class ActiveMediaList
 
     const std::unordered_set<size_t> getActiveTalkers() const;
 
+    utils::Optional<uint32_t> getUserId(const uint32_t ssrc);
```
olofkallander, line 73: remove

api/Parser.cpp
reddvl1980: removed.

bridge/Mixer.cpp
```diff
@@ -803,12 +804,29 @@ bool Mixer::getEndpointExtendedInfo(const std::string& endpointId,
...
+        endpoint.ssrcRewritten = audio->second->localSsrc;
+    }
+
+    auto remote = transport->getRemotePeer();
+    endpoint.remotePort = remote.getPort();
+    endpoint.remoteIP = remote.ipToString();
+    endpoint.localIP = transport->getLocalRtpPort().ipToString();
+    endpoint.localPort = transport->getLocalRtpPort().getPort();
+    auto transportType = transport->getSelectedTransportType();
+    endpoint.protocol = (transportType.isSet()) ? ice::toString(transportType.get()) : "n/a";
```
olofkallander, line 828: move third ) to the end

bridge/engine/ActiveMediaList.cpp
reddvl1980: Fixed.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -39,6 +39,7 @@ ActiveMediaList::ActiveMediaList(size_t instanceId,
       _maxSpeakers(audioSsrcs.size()),
       _audioParticipants(maxParticipants),
       _incomingAudioLevels(32768),
+      _incomingPttEvents(32768),
```
olofkallander, line 42: I pondered this again. Perhaps it is better to add a bool ptt to AudioLevelEntry, since we report that one always from Transport. Then we need only one queue and one loop to process the audio levels and push the noiseLevel to desired value.

reddvl1980: Good. Going to fix that.

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -160,12 +166,16 @@ void AudioForwarderReceiveJob::run()
             silence = !isPtt.get();
             if (!audioLevel.isSet())
             {
-                audioLevel.set(silence ? 127 : 0);
+                // We need to 'fake' reasonable audio levels for the reasons.
```
olofkallander, line 169: We can remove the comment. If is obivous we send in fixed audio levels.

olofkallander (approved):

PR #122, opened 2022-08-02
media fwd barbell

olofkallander:


memory/StackMap.h
```diff
@@ -0,0 +1,302 @@
...
+            {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    size_t capacity() const { return _elements.size(); }
+    size_t size() const { return _count; }
+    bool empty() const { return _count == 0; }
```
RicardoMDomingues, line 212: It must remove all `commited` flags and `indexEntry.position`.

memory/StackMap.h
```diff
@@ -0,0 +1,302 @@
...
+
+        return 0;
+    }
+
+    std::array<IndexEntry, SIZE * 4> _index;
+    std::array<ElementEntry, SIZE> _elements;
+    const uint32_t _end;
+    uint32_t _nextFreeEntry = 0;
+    T _emptyObject;
+    uint32_t _maxSpread;
```
RicardoMDomingues, line 299: `_maxSpread` on increase never, decrease. One option would be clean the `spread` on `clean` but I guess call `clean` function will be rare cases.

memory/StackMap.h
```diff
@@ -0,0 +1,302 @@
...
+    std::pair<iterator, bool> add(const KeyT& key, const T& value)
+    {
+        auto keyHash = utils::hash<KeyT>{}(key);
+
+        for (uint32_t i = 0; i < _maxSpread; ++i)
+        {
+            const auto& indexEntry = _index[indexPosition(keyHash, i)];
+            if (indexEntry.keyHash == keyHash && indexEntry.position)
+            {
+                return std::make_pair(end(), false); // no duplets
```
RicardoMDomingues, line 89: This will prevent add elements with hash collisions as well. But there is a simple way to solve that that (we will need twice the memory to `_index`and rerange that after an an X number of erases, but we could augment the load factor to reduce the memory, 25% sounds a bit exaggerated)

memory/StackArray.h
```diff
@@ -7,9 +7,12 @@ template <typename T, size_t SIZE>
 class StackArray
 {
 public:
-    explicit StackArray(size_t size) : _size(size)
+    typedef const T* const_iterator;
+    typedef T* iterator;
+
+    StackArray() : _capacity(SIZE), _size(0), _dataPtr(_data) {}
```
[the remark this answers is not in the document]
olofkallander: Yes, it is a bit like std:array but with ability to be instantiated with dynamic memory if needed.
Originally we used it when handling inbound sctp messages, that are usually small enough to be handled on stack within a certain limit. If it would be bigger occasionally, the heap is used.

memory/StackArray.h
```diff
@@ -7,9 +7,12 @@ template <typename T, size_t SIZE>
...
+    explicit StackArray(size_t size) : _capacity(size), _size(0), _dataPtr(_data)
```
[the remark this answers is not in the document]
olofkallander: I actually am not too proud about the name. Stack gives the impression it is a stack but it refers only to how it is stored. It is an Array.

memory/StackArray.h
```diff
@@ -27,17 +30,44 @@ class StackArray
...
     T& operator[](size_t pos) { return _dataPtr[pos]; }
     const T& operator[](size_t pos) const { return _dataPtr[pos]; }
 
     const T* data() const { return _dataPtr; }
-    T* data() { return _dataPtr; }
+    StackArray& append(const T* vector, size_t count)
+    {
+        if (_size + count <= _capacity)
+        {
+            std::memcpy((_dataPtr + _size), vector, count * sizeof(T));
```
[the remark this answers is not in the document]
olofkallander: We should do copy in this case. I will change it.

memory/StackMap.h
```diff
@@ -0,0 +1,302 @@
...
+
+    const_iterator cend() const { return const_iterator(const_cast<ElementEntry*>(_elements.data()), _end, _end); }
+
+    const_iterator begin() const { return cbegin(); }
+    const_iterator end() const { return cend(); }
+    iterator begin() { return iterator(cbegin()); }
+    iterator end() { return iterator(_elements.data(), _end, _end); }
+
+    template <typename TypeName = T>
+    TypeName getItem(const KeyT& key, typename std::enable_if<std::is_pointer<TypeName>::value>::type* = nullptr)
```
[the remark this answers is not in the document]
olofkallander: I tried that but it couldn't differentiate between the 2 getItem overloads, as it is not a parameter. That is my understanding of the compile errors. Perhaps I missed the correct way to do it, but I could also not find a working example where it is not used as param.

memory/StackMap.h
olofkallander: thanks, it should be reset.

memory/StackMap.h
olofkallander: I know, that is the weakness of this simple solution. We could store the original key, compare on that and decide to find another slot to the right and increase _maxSpread.

olofkallander: It is the same with MpmcHashMap. The key is limited to 40 bits in that case and it is necessary to allow atomic read and update of the key and position. But it means a hash collision prevents you from adding that object.
Comparing the original keys would render a sequence of atomic reads, and double reads that makes it impractical to use. And much harder to implement.

olofkallander: In this case with 64 bits, hash collision is very rare unless the hash function is poor.

memory/StackMap.h
olofkallander: you are thinking about clear method

memory/StackMap.h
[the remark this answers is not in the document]
RicardoMDomingues: Perhaps in your try you forgot to reverse the enable condition. If both are enable at the same time then the signature will be the same and will get a compile error. When one is enabled, the other must be disabled,  like this:

```
std::enable_if_t<std::is_pointer<TypeName>, TypeName>
getItem(const KeyT& key)
{
   return it->second;
}


std::enable_if_t<!std::is_pointer<TypeName>, TypeName>
getItem(const KeyT& key)
{
    return &it->second;
}
```

There is also another option to apply the enable condition on iterator so you don't need to implement `getItem` two times. I will open a PR with the suggestion

memory/StackMap.h
RicardoMDomingues: Yes, wrong line :)

RicardoMDomingues: Yeah sorry, never mind. I guess I misinterpreted the code as `_count = 0` and my brain auto renamed this method to "clean". I am seeing now there is no clean method implemented. You can forget the comment

PR #125, opened 2022-08-05
API video ssrc change

olofkallander:


api/EndpointDescription.h
```diff
@@ -91,14 +89,26 @@ struct EndpointDescription
...
     {
-        utils::Optional<Transport> _transport;
-
-        std::vector<uint32_t> _ssrcs;
-        std::vector<SsrcGroup> _ssrcGroups;
-        std::vector<PayloadType> _payloadTypes;
-        std::vector<std::pair<uint32_t, std::string>> _rtpHeaderExtensions;
-        utils::Optional<std::vector<uint32_t>> _ssrcWhitelist;
-        std::vector<SsrcAttribute> _ssrcAttributes;
+        utils::Optional<Transport> transport;
```
marcusspangenberg, line 92: I think you forgot the member prefix _

marcusspangenberg:
Is there a change in the API related to this? Will you update the examples/docs as well?

olofkallander:
> Is there a change in the API related to this? Will you update the examples/docs as well? 

Yes, they should be updated already.

RicardoMDomingues (approved):
LGTM, @olofkallander I made some fixes. Please take a look

api/Parser.cpp
```diff
@@ -357,53 +357,40 @@ EndpointDescription parsePatchEndpoint(const nlohmann::json& data, const std::st
...
+            videoChannel.streams.emplace_back();
+            auto& videoStream = videoChannel.streams.back();
+            for (const auto& rtpSource : requiredJsonArray(stream, "sources"))
             {
-                const auto ssrc =
-                    ssrcJson.is_string() ? std::stoul(ssrcJson.get<std::string>()) : ssrcJson.get<uint32_t>();
-                ssrcAttribute._ssrcs.push_back(ssrc);
+                api::EndpointDescription::VideoStream::Level level;
+                const auto& main = rtpSource["main"];
+                level.main = (main.is_string() ? std::stoul(main.get<std::string>()) : main.get<uint32_t>());
```
olofkallander, line 385: Should we really support this. We specify the API and there is no point in doing ssrc as string, and no examples of it.

PR #126, opened 2022-08-06
Simplify enable_if

RicardoMDomingues:


test/memory/MapTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    ASSERT_EQ(nullptr, map1.getItem(10));
+    ASSERT_EQ(nullptr, map2.getItem(20));
+}
+
+TEST(Map, getItemShouldReturnPointerToConst)
+{
+    // As we use initialized memory, const types are not supported as
+    // then internal memory of Map is immutable
+    //auto map1 = Map<int, const int, 256>();
```
RicardoMDomingues, line 89: `const types`  are not supported because we are using initialised memory then it's immutable memory when we have const types.

Perhaps we will not have any use case of this. It's still support `const T*` as the T is mutable pointer for an immutable type. Then T itself is mutable (which is different of having immutable pointers, like `T* const`).

olofkallander (approved):

PR #127, opened 2022-08-08
Reduce log spam on request logs

RicardoMDomingues:


bridge/RequestLogger.h
```diff
@@ -1,27 +1,39 @@
...
 #include <cstdint>
 #include <string>
+#include <vector>
 
 namespace bridge
 {
 
 class RequestLogger
 {
+    const static std::vector<std::string> kLogFilter;
```
olofkallander, line 18: just logFilter

olofkallander:
I just wonder. It feels like it may have been a mistake to leave to leave the request logging for all requests in there. At info level is it?
It sounds like a debugging measure that perhaps was intended to be temporary.

RicardoMDomingues:
> I just wonder. It feels like it may have been a mistake to leave to leave the request logging for all requests in there. At info level is it? It sounds like a debugging measure that perhaps was intended to be temporary.

This was consequence of this refactor #113. Before we sent the response to "about" and "stats" endpoint before create `RequestLogger` on `ApiRequestHandler::onRequest`. Now we still create the `RequestLogger` on  `ApiRequestHandler::onRequest` but we then we call "`callEndpointAction` to handle everything,  so `RequestLogger` is logging all requests

reddvl1980 (approved):

olofkallander (approved):

PR #129, opened 2022-08-10
breakout fixes

olofkallander:
Made a break out of fixes to reduce size of bb branch. Hopefully it will be easier to rebase on this.


concurrency/MpmcHashmap.cpp
```diff
@@ -51,14 +51,6 @@ bool MurmurHashIndex::add(uint64_t key, uint32_t value)
         }
     }
 
-    int emptySlots = 0;
-    for (auto& entry : _index)
```
reddvl1980, line 55: Just to clarify: it used to track being used keys, recover them and fail if nothing was recovered. But now it no recovery attempt is being made?

concurrency/MpmcHashmap.h
```diff
@@ -14,6 +17,8 @@ namespace concurrency
 // No value may be zero !!
 // key is 40bits and will be truncated to 40 bits.
 // elementCount must be power of two and <= 16777216
+// Since key is prehashed and not stored in the index, there is a risk of hash collection and you will only notice by
```
reddvl1980, line 20: Typo "collision" instead of "collection.

reddvl1980 (approved):
LGTM with one typo.

memory/Array.h
```diff
@@ -0,0 +1,86 @@
...
+public:
+    typedef const T* const_iterator;
+    typedef T* iterator;
+
+    Array() : _capacity(SIZE), _size(0), _dataPtr(reinterpret_cast<T*>(_data)) {}
+    explicit Array(size_t size) : _capacity(size), _size(0), _dataPtr(reinterpret_cast<T*>(_data))
+    {
+        if (size > SIZE)
+        {
+            _dataPtr = reinterpret_cast<T*>(malloc(size * sizeof(T)));
```
RicardoMDomingues, line 18: [minor] malloc will allocate memory aligned to `std::max_align_t` (which I guess is 16 byte) which should work well for all our cases. But we have a better function to allocate this with alignment specification so we don't need a strict align to 16 when we are storing bytes.

`std::aligned_alloc(std::alignment_of<T>::value, size * sizeof(T))`

RicardoMDomingues (approved):
LGTM, only minor comment

PR #130, opened 2022-08-11
Add static ICE endpoints for probing STUN requests

stepanaksenov:
So far I added new ProbeServer class which implements listener interfaces for both UdpEndpoint and TcpServerEndpoint. It will handle incoming probes and reply with OK STUN response.

Current questions/concerns:

- Ownership of "probing endpoints". Such simplified (protocol/IP/port) ICE candidates should be based on a shared (UDP) endpoint and TcpServerEndpoint. Currently the endpoints are created/owned by TransportFactory. It should either take on responsibility to build the probing endpoints and expose them or share actual endpoints with new ProbeServer.
- Closing TCP connection after replying to probing STUN request. As far as I understand we currently close TCP connection in these cases: 1) invalid STUN request 2) attack 3) connection closed by client. TcpEndpoint owns file descriptor and closes it when required. There is no mechanism to close connection from TCP endpoint listener as for now.
- Ownership of static ICE credentials. The credentials should be used to register listener on endpoints and also shared via API with MBR/CS. It seems more appropriate if ProbeServer generates and exposes them. However, it would mean that probing functionality is split between ProbeServer (credentials) and TransportFactory (probing endpoints), which doesn't seem good.

### DIFF
--- a/transport/TransportFactory.cpp
+++ b/transport/TransportFactory.cpp
@@ -2,6 +2,7 @@
 #include "concurrency/MpmcHashmap.h"
 #include "config/Config.h"
 #include "memory/PacketPoolAllocator.h"
+#include "transport/ProbeServer.h"
 #include "transport/RecordingTransport.h"
 #include "transport/RtcTransport.h"
 #include "transport/TcpEndpoint.h"
@@ -75,6 +76,9 @@ public:
         {
             for (uint32_t portOffset = 0; portOffset < std::max(1u, config.ice.sharedPorts.get()); ++portOffset)
             {
+                // Generate ufrag pass
+                // IceSession staticIceSession(ufrag, pass)
+                // Ip:portx`
                 _sharedEndpoints.push_back(Endpoints());
                 for (SocketAddress portAddress : interfaces)
                 {
@@ -482,6 +486,36 @@ public:
         logger::debug("closing %s", _name, endpoint->getName());
         ++_pendingTasks;
         endpoint->stop(this);
+    }
+
+    void registerProbeServer(ProbeServer& server) override
+    {
+        for (auto& endpoint : _tcpServerEndpoints)
+        {
+            server.registerEndpoint(*endpoint);
+        }
+
+        auto udpProbeEndpoint = _sharedEndpoints[0]; // config.ice.singlePort
+
+        for (auto& endpoint : udpProbeEndpoint)
+        {
+            server.registerEndpoint(*endpoint);
+        }  
+    }
+
+    void unregisterProbeServer(ProbeServer& server) override
+    {
+        for (auto& endpoint : _tcpServerEndpoints)
+        {
+            server.unregisterEndpoint(*endpoint);
+        }
+
+        auto udpProbeEndpoint = _sharedEndpoints[0]; // config.ice.singlePort
+
+        for (auto& endpoint : udpProbeEndpoint)
+        {
+            server.unregisterEndpoint(*endpoint);
+        }
     }
 
 private:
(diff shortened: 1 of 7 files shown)


transport/TransportFactory.cpp
```diff
@@ -484,6 +488,36 @@ class TransportFactoryImpl final : public TransportFactory,
...
     }
 
+    void registerProbeServer(ProbeServer& server) override
+    {
+        for (auto& endpoint : _tcpServerEndpoints)
+        {
+            server.registerEndpoint(*endpoint);
+        }
+
+        auto udpProbeEndpoint = _sharedEndpoints[0]; // config.ice.singlePort
```
RicardoMDomingues, line 498: register in all interfaces on _sharedEndpoints. On current state we will have only one but we already discuss in having a range of ports as we are reaching the limits the packets we can send in one single port on load tests

RicardoMDomingues: In the end if the port range it big, we will have a lot of ICE candidates as well, but we can filter some of them and only provide 2 or 3 of them

stepanaksenov: I choose one UDP endpoint because of @olofkallander line in the ticket "think one of the shared UDP ports we have could be used to handle this". I can change it to add listener (ProbeServer) to all the UDP endpoints.

stepanaksenov: Now I think it'd be better to simply pass all the TcpServer and UDP endpoints from TransportFactory to ProbeServer in register method and let the server decide which ones to use for probing.

```
register(ProbeServer& server)
{
    server.attach(_sharedEndpoints, _tcpServerEndpoints);
}
```

RicardoMDomingues: Yeah, you can do like that but keep in mind probably we want to have all endpoints for the same port. Imagine you have 2 NIC. 1st IP `10.0.0.1` and 2nd `10.0.0.2`. And we have port range from 10000-10005.

So even if we want to select only one port (let's say port 10000) you should include both endpoints `10.0.0.1:1000` and `10.0.0.2:1000` because we don't know which IP is reachable from client. That perhaps is more difficult to control inside on `ProbeServer`. `TransportFactory` has a vector with another vector. First vector is about the port and the second vector is all network interfaces where that port is bound.

I think if one day we will have more than 1 port available, maybe we should provide candidates for at least for 2 different ports, so in a load situation where 1 port is very loaded and some packets will be lost, client can use a second port to probe

transport/TransportFactory.h
```diff
@@ -67,6 +68,12 @@ class TransportFactory
     virtual bool openRtpMuxPorts(Endpoints& rtpPorts) const = 0;
 
     virtual void maintenance(uint64_t timestamp) = 0;
+
+    virtual void registerProbeServer(ProbeServer& server) = 0;
```
olofkallander, line 72: You can fix it later, but I think TransportFactory can be unaware of the ProbeServer type. 
Just take the Endpoint::IEvents and ServerEndpoint::IEvents interfaces. Perhaps even two register methods, one for each interface.

stepanaksenov: We need to provide ICE credentials when registering a listener for endpoints, as `registerListener(const std::string& stunUserName, IEvents* listener)`. I opted to have them in ProbeServer, and let it accept endpoints for registration, hence TransportFactory knows about the ProbeServer.

I can change it and move static ICE credentials to TransportFactory also.

transport/TransportFactory.cpp
stepanaksenov: Yes, here _sharedEndpoints is vector of vectors

```
std::vector<Endpoints>
typedef std::vector<std::shared_ptr<Endpoint>> Endpoints;
```

Also as I understand, we need to **listen** on all the interfaces, but in API we report only public IPv4 and optionally IPv6 addresses. So ProbeServer would only need to know port of specific endpoint and retrieve public IPs from ICE config.

PR #131, opened 2022-08-15
Add static ICE endpoints

stepanaksenov:
Added new API endpoint "probing", which returns list of ICE endpoints so that clients could perform RTT measurements.
Each endpoint expects STUN Binding Request with provided ICE credentials and replies with  "ok" STUN response.
Static ICE credentials are generated on every SMB startup.

RTC-13083

### DIFF
--- a/api/Parser.cpp
+++ b/api/Parser.cpp
@@ -162,29 +162,7 @@ api::EndpointDescription::Transport parsePatchEndpointTransport(const nlohmann::
 
     if (data.find("ice") != data.end())
     {
-        const auto& iceJson = data["ice"];
-        api::EndpointDescription::Ice ice;
-        ice._ufrag = iceJson["ufrag"].get<std::string>();
-        ice._pwd = iceJson["pwd"].get<std::string>();
-
-        for (const auto& candidateJson : optionalJsonArray(iceJson, "candidates"))
-        {
-            api::EndpointDescription::Candidate candidate;
-            candidate._generation = candidateJson["generation"].get<uint32_t>();
-            candidate._component = candidateJson["component"].get<uint32_t>();
-            candidate._protocol = candidateJson["protocol"].get<std::string>();
-            candidate._port = candidateJson["port"].get<uint32_t>();
-            candidate._ip = candidateJson["ip"].get<std::string>();
-            setIfExists(candidate._relPort, candidateJson, "rel-port");
-            setIfExists(candidate._relAddr, candidateJson, "rel-addr");
-            candidate._foundation = candidateJson["foundation"].get<std::string>();
-            candidate._priority = candidateJson["priority"].get<uint32_t>();
-            candidate._type = candidateJson["type"].get<std::string>();
-            candidate._network =
-                candidateJson.find("network") != candidateJson.end() ? candidateJson["network"].get<uint32_t>() : 0;
-            ice._candidates.emplace_back(std::move(candidate));
-        }
-
+        api::EndpointDescription::Ice ice = api::Parser::parseIce(data["ice"]);
         transport._ice.set(ice);
     }
 
@@ -519,6 +497,32 @@ ConferenceEndpointExtendedInfo parseEndpointExtendedInfo(const nlohmann::json& d
     return endpoint;
 }
 
+EndpointDescription::Ice parseIce(const nlohmann::json& iceJson)
+{
+    api::EndpointDescription::Ice ice;
+    ice._ufrag = iceJson["ufrag"].get<std::string>();
+    ice._pwd = iceJson["pwd"].get<std::string>();
+
+    for (const auto& candidateJson : optionalJsonArray(iceJson, "candidates"))
+    {
+        api::EndpointDescription::Candidate candidate;
+        candidate._generation = candidateJson["generation"].get<uint32_t>();
+        candidate._component = candidateJson["component"].get<uint32_t>();
+        candidate._protocol = candidateJson["protocol"].get<std::string>();
+        candidate._port = candidateJson["port"].get<uint32_t>();
+        candidate._ip = candidateJson["ip"].get<std::string>();
+        setIfExists(candidate._relPort, candidateJson, "rel-port");
+        setIfExists(candidate._relAddr, candidateJson, "rel-addr");
+        candidate._foundation = candidateJson["foundation"].get<std::string>();
+        candidate._priority = candidateJson["priority"].get<uint32_t>();
+        candidate._type = candidateJson["type"].get<std::string>();
+        candidate._network =
+            candidateJson.find("network") != candidateJson.end() ? candidateJson["network"].get<uint32_t>() : 0;
+        ice._candidates.emplace_back(std::move(candidate));
+    }
+
+    return ice;
+}
 } // namespace Parser
 
 } // namespace api
--- a/bridge/ApiRequestHandler.cpp
+++ b/bridge/ApiRequestHandler.cpp
@@ -23,8 +23,10 @@ httpd::Response makeErrorResponse(httpd::StatusCode statusCode, const std::strin
 namespace bridge
 {
 
-ApiRequestHandler::ApiRequestHandler(bridge::MixerManager& mixerManager, transport::SslDtls& sslDtls)
-    : ActionContext(mixerManager, sslDtls),
+ApiRequestHandler::ApiRequestHandler(bridge::MixerManager& mixerManager,
+    transport::SslDtls& sslDtls,
+    transport::ProbeServer& probeServer)
+    : ActionContext(mixerManager, sslDtls, probeServer),
       _lastAutoRequestId(0)
 
 #if ENABLE_LEGACY_API
@@ -70,6 +72,9 @@ httpd::Response ApiRequestHandler::callEndpointAction(RequestLogger& requestLogg
     if (utils::StringTokenizer::isEqual(token, "barbell") && token.next &&
         (request._method == httpd::Method::POST || request._method == httpd::Method::DELETE))
         return processBarbellAction(this, requestLogger, request, token);
+
+    if (utils::StringTokenizer::isEqual(token, "ice-candidates") && request._method == httpd::Method::GET)
+        return getProbingInfo(this, requestLogger, request, token);
 
     throw httpd::RequestErrorException(httpd::StatusCode::METHOD_NOT_ALLOWED,
         utils::format("HTTP method '%s' not allowed on this endpoint", request._methodString.c_str()));
--- a/bridge/Bridge.cpp
+++ b/bridge/Bridge.cpp
@@ -6,7 +6,9 @@
 #include "httpd/Httpd.h"
 #include "jobmanager/JobManager.h"
 #include "jobmanager/WorkerThread.h"
+#include "transport/Endpoint.h"
 #include "transport/EndpointFactoryImpl.h"
+#include "transport/ProbeServer.h"
 #include "transport/RtcePoll.h"
 #include "transport/TransportFactory.h"
 #include "transport/dtls/SrtpClientFactory.h"
@@ -82,6 +84,11 @@ Bridge::~Bridge()
         _jobManager->stop();
     }
     logger::info("JobManager stopped", "main");
+
+    if (_probeServer)
+    {
+        _probeServer->stop();
+    }
 
     uint32_t n = 0;
     for (auto& workerThread : _workerThreads)
@@ -159,6 +166,16 @@ void Bridge::initialize(std::shared_ptr<transport::EndpointFactory> endpointFact
         return;
     }
 
+    _probeServer = std::make_unique<transport::ProbeServer>(_iceConfig, _config);
+
+    const auto credentials = _probeServer->getCredentials();
+
+    _transportFactory->registerIceListener(*static_cast<transport::Endpoint::IEvents*>(_probeServer.get()),
+        credentials.first);
+
+    _transportFactory->registerIceListener(*static_cast<transport::ServerEndpoint::IEvents*>(_probeServer.get()),
+        credentials.first);
+
     _mixerManager = std::make_unique<bridge::MixerManager>(*_idGenerator,
         *_ssrcGenerator,
         *_jobManager,
@@ -169,7 +186,7 @@ void Bridge::initialize(std::shared_ptr<transport::EndpointFactory> endpointFact
         *_sendPacketAllocator,
         *_audioPacketAllocator);
 
-    _requestHandler = std::make_unique<bridge::ApiRequestHandler>(*_mixerManager, *_sslDtls);
+    _requestHandler = std::make_unique<bridge::ApiRequestHandler>(*_mixerManager, *_sslDtls, *_probeServer);
 
     const auto httpAddress = transport::SocketAddress::parse(_config.address, _config.port);
     _httpd = std::make_unique<httpd::Httpd>(*_requestHandler);
--- a/bridge/endpointActions/GetProbingInfo.cpp
+++ b/bridge/endpointActions/GetProbingInfo.cpp
@@ -1,0 +1,63 @@
+#include "ApiActions.h"
+#include "ApiHelpers.h"
+#include "api/ConferenceEndpoint.h"
+#include "bridge/Mixer.h"
+#include "bridge/MixerManager.h"
+#include "bridge/RequestLogger.h"
+#include "nlohmann/json.hpp"
+#include "transport/ProbeServer.h"
+
+namespace bridge
+{
+httpd::Response getProbingInfo(ActionContext* context,
+    RequestLogger& requestLogger,
+    const httpd::Request&,
+    const utils::StringTokenizer::Token&)
+{
+    nlohmann::json responseBodyJson;
+    nlohmann::json iceJson;
+
+    auto& probeServer = context->probeServer;
+
+    const auto credentials = probeServer.getCredentials();
+
+    iceJson["ufrag"] = credentials.first;
+    iceJson["pwd"] = credentials.second;
+
+    nlohmann::json candidatesJson = nlohmann::json::array();
+
+    for (const auto& candidate : probeServer.getCandidates())
+    {
+        api::EndpointDescription::Candidate description = iceCandidateToApi(candidate);
+
+        nlohmann::json candidateJson;
+        candidateJson["generation"] = description._generation;
+        candidateJson["component"] = description._component;
+        candidateJson["protocol"] = description._protocol;
+        candidateJson["port"] = description._port;
+        candidateJson["ip"] = description._ip;
+        candidateJson["foundation"] = description._foundation;
+        candidateJson["priority"] = description._priority;
+        candidateJson["type"] = description._type;
+        candidateJson["network"] = description._network;
+        if (description._relAddr.isSet())
+        {
+            candidateJson["rel-addr"] = description._relAddr.get();
+        }
+        if (description._relPort.isSet())
+        {
+            candidateJson["rel-port"] = description._relPort.get();
+        }
+
+        candidatesJson.push_back(candidateJson);
+    }
+
+    iceJson["candidates"] = candidatesJson;
+    responseBodyJson["ice"] = iceJson;
+
+    httpd::Response response(httpd::StatusCode::OK, responseBodyJson.dump(4));
+    response._headers["Content-type"] = "text/json";
+    requestLogger.setResponse(response);
+    return response;
+}
+} // namespace bridge
--- a/transport/ProbeServer.cpp
+++ b/transport/ProbeServer.cpp
@@ -1,0 +1,368 @@
+#include "transport/ProbeServer.h"
+#include "concurrency/ThreadUtils.h"
+#include "ice/IceCandidate.h"
+#include "logger/Logger.h"
+#include "transport/ice/IceSession.h"
+#include "transport/ice/Stun.h"
+#include "utils/ContainerAlgorithms.h"
+#include "utils/Time.h"
+#include <unistd.h>
+
+namespace transport
+{
+ProbeServer::ProbeServer(const ice::IceConfig& iceConfig, const config::Config& config)
+    : _iceConfig(iceConfig),
+      _config(config),
+      _queue(1024),
+      _maintenanceThreadIsRunning(true),
+      _maintenanceThread(new std::thread([this] { this->run(); }))
+{
+    char ufrag[14 + 1];
+    char pwd[24 + 1]; // length selected to make attribute *4 length
+
+    ice::IceSession::generateCredentialString(_idGenerator, ufrag, sizeof(ufrag) - 1);
+    ice::IceSession::generateCredentialString(_idGenerator, pwd, sizeof(pwd) - 1);
+
+    _credentials = std::make_pair<std::string, std::string>(ufrag, pwd);
+}
+
+// Endpoint::IEvents
+void ProbeServer::onRtpReceived(Endpoint& endpoint,
+    const SocketAddress& source,
+    const SocketAddress& target,
+    memory::UniquePacket packet)
+{
+}
+
+void ProbeServer::onDtlsReceived(Endpoint& endpoint,
+    const SocketAddress& source,
+    const SocketAddress& target,
+    memory::UniquePacket packet){};
+
+void ProbeServer::onRtcpReceived(Endpoint& endpoint,
+    const SocketAddress& source,
+    const SocketAddress& target,
+    memory::UniquePacket packet)
+{
+}
+
+void ProbeServer::onIceReceived(Endpoint& endpoint,
+    const SocketAddress& source,
+    const SocketAddress& target,
+    memory::UniquePacket packet)
+{
+    replyStunOk(endpoint, source, std::move(packet));
+}
+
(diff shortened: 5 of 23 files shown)


bridge/Bridge.cpp
```diff
@@ -152,6 +154,16 @@ void Bridge::initialize()
         return;
     }
 
+    _probeServer = std::make_unique<transport::ProbeServer>(_iceConfig);
+
+    const auto credentials = _probeServer->getProbeCredentials();
+
+    _transportFactory->registerIceListener(*static_cast<transport::Endpoint::IEvents*>(_probeServer.get()),
```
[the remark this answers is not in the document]
stepanaksenov: It is both Endpoint::IEvents and ServerEndpoint::IEvents, without the cast compiler can't resolve which method to use on TransportFactory.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+    return _probeCredentials;
+}
+
+const std::vector<ProbeCandidate> ProbeServer::getProbeCandidates()
+{
+    const std::lock_guard<std::mutex> locker(_endpointsLock);
+    return _probeCandidates;
+}
+
+// TODO copied from IceSession.cpp, move to utils/tools and reuse;
```
olofkallander, line 199: We should make it static in IceSession and let it take the random token as argument.

stepanaksenov: The method generates new tokens for each 10th element (_idGenerator.next()). I guess we should StunTransactionIdGenerator then.

bridge/ApiRequestHandler.cpp
```diff
@@ -71,6 +73,9 @@ httpd::Response ApiRequestHandler::callEndpointAction(RequestLogger& requestLogg
         (request._method == httpd::Method::POST || request._method == httpd::Method::DELETE))
         return processBarbellAction(this, requestLogger, request, token);
 
+    if (utils::StringTokenizer::isEqual(token, "probing") && request._method == httpd::Method::GET)
```
olofkallander, line 76: Can we find a better name for the endpoint? The purpose is to get ice credentials for making a STUN RTT probe.

transport/TransportFactory.cpp
```diff
@@ -460,7 +460,10 @@ class TransportFactoryImpl final : public TransportFactory,
         return metrics;
     }
 
-    bool isGood() const override { return _good; }
+    bool isGood() const override
```
olofkallander, line 463: what settings do you have for clang-format ?

bridge/ApiRequestHandler.cpp
stepanaksenov: Simply "probecandidates" maybe?

transport/TransportFactory.cpp
stepanaksenov: Wrong ones, I guess. I could not run clang-format and used C/C++ extension instead. I will fix clang-format and reformat the code.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+        auto address = _iceConfig.publicIpv4;
+        address.setPort(port);
+        _probeCandidates.emplace_back(ProbeCandidate{ice::TransportType::UDP, address});
+
+        _udpPorts.insert(port);
+
+        logger::debug("Registered UDP endpoint %s:%u", _name, address.ipToString().c_str(), port);
+    }
+
+    if (!_iceConfig.publicIpv6.empty() && endpoint.getLocalPort().getFamily() == AF_INET6)
```
olofkallander, line 76: The ipv6 is always public so I wonder if this reference to config is necessary.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+        _tcpPorts.insert(port);
+
+        logger::debug("Registered TCP server endpoint %s:%u", _name, address.ipToString().c_str(), port);
+    }
+
+    if (!_iceConfig.publicIpv6.empty() && endpoint.getLocalPort().getFamily() == AF_INET6)
+    {
+        auto address = _iceConfig.publicIpv6;
+        address.setPort(port);
+        _probeCandidates.emplace_back(ProbeCandidate{ice::TransportType::TCP, address});
```
olofkallander, line 130: I think emplace_back can even take the constructor arguments directly for the ProbeCandidate.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+    const SocketAddress& target,
+    memory::UniquePacket packet)
+{
+    if (endpoint->getTransportType() == ice::TransportType::TCP)
+    {
+        replyStunOk(*endpoint, source, std::move(packet));
+
+        // wait for while to ensure the reply is sent,
+        // before shutting down endpoint
+        u_int32_t gracePeriodMs = 1000;
```
olofkallander, line 156: We cannot do this here as it will hog a RT worker thread. It also creates an attack vector on SMB.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+    const SocketAddress& source,
+    const SocketAddress& target,
+    memory::UniquePacket packet)
+{
+    replyStunOk(endpoint, source, std::move(packet));
+}
+
+void ProbeServer::onRegistered(Endpoint& endpoint)
+{
+    const std::lock_guard<std::mutex> locker(_endpointsLock);
```
olofkallander, line 51: I guess we can have it here since it is calle donce at startup, but this is one of the RT workers and they may not pass a mutex at any time.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+// ICE
+const std::pair<std::string, std::string>& ProbeServer::getProbeCredentials() const
+{
+    return _probeCredentials;
+}
+
+const std::vector<ProbeCandidate> ProbeServer::getProbeCandidates()
+{
+    const std::lock_guard<std::mutex> locker(_endpointsLock);
```
olofkallander, line 195: Once the candidates have been setup they are const here and no lock should be needed. I wonder even if it is a good idea to accept any API requests before we have opened ports and registered the probe server.
The right way is to have no locks in ProbeServer and instead make sure we do not listen to http until probeserver has fully registered.

bridge/ApiRequestHandler.cpp
olofkallander: or ice-candidates. That is what we are retrieving.

bridge/endpointActions/GetProbingInfo.cpp
```diff
@@ -0,0 +1,45 @@
...
+        candidateJson["ip"] = candidate.address.ipToString();
+        candidateJson["port"] = candidate.address.getPort();
+        candidatesJson.push_back(candidateJson);
+    }
+
+    iceJson["candidates"] = candidatesJson;
+    responseBodyJson["ice"] = iceJson;
+
+    httpd::Response response(httpd::StatusCode::OK, responseBodyJson.dump(4));
+    response._headers["Content-type"] = "text/json";
```
olofkallander, line 41: I think we should stick to regular letters.

bridge/ApiRequestHandler.cpp
stepanaksenov: I tried to distinguish between real ICE candidates and "simplified" ones, intended for probing only.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+void ProbeServer::replyStunOk(Endpoint& endpoint, const SocketAddress& destination, memory::UniquePacket packet)
+{
+    uint64_t timestamp = utils::Time::getAbsoluteTime();
+    const void* data = packet->get();
+
+    auto* stunMessage = ice::StunMessage::fromPtr(data);
+
+    if (stunMessage && stunMessage->isValid() && stunMessage->header.isRequest() &&
+        stunMessage->isAuthentic(_probeCredentials.second))
```
RicardoMDomingues, line 174: Not sure if this verification fails and is a TCP we want to close sockets as soon as possible

transport/TransportFactory.cpp
```diff
@@ -484,6 +487,44 @@ class TransportFactoryImpl final : public TransportFactory,
         endpoint->stop(this);
     }
 
+    void registerIceListener(Endpoint::IEvents& listener, const std::string& ufrag) override
+    {
+        for (auto endpoints : _sharedEndpoints)
```
RicardoMDomingues, line 492: Avoid vector copy

```suggestion
        for (auto& endpoints : _sharedEndpoints)
```

transport/TransportFactory.cpp
```diff
@@ -484,6 +487,44 @@ class TransportFactoryImpl final : public TransportFactory,
...
+        {
+            for (auto endpoint : endpoints)
```
RicardoMDomingues, line 494: Less important here but avoid a `shared_ptr` will save some atomic operations on ref count
```suggestion
            for (auto& endpoint : endpoints)
```

transport/TransportFactory.cpp
```diff
@@ -484,6 +487,44 @@ class TransportFactoryImpl final : public TransportFactory,
...
+            {
+                endpoint->registerListener(ufrag, &listener);
+            }
+        }
+    }
+
+    void registerIceListener(ServerEndpoint::IEvents& listener, const std::string& ufrag) override
+    {
+        for (auto endpoint : _tcpServerEndpoints)
```
RicardoMDomingues, line 503:
```suggestion
        for (auto& endpoint : _tcpServerEndpoints)
```

transport/TransportFactory.cpp
```diff
@@ -484,6 +487,44 @@ class TransportFactoryImpl final : public TransportFactory,
...
+        {
+            endpoint->registerListener(ufrag, &listener);
+        }
+    }
+
+    void unregisterIceListener(Endpoint::IEvents& listener, const std::string& ufrag) override
+    {
+        for (auto endpoints : _sharedEndpoints)
```
RicardoMDomingues, line 511:
```suggestion
        for (auto& endpoints : _sharedEndpoints)
```

transport/TransportFactory.cpp
```diff
@@ -484,6 +487,44 @@ class TransportFactoryImpl final : public TransportFactory,
...
+        {
+            for (auto endpoint : endpoints)
```
RicardoMDomingues, line 513:
```suggestion
            for (auto& endpoint : endpoints)
```

transport/TransportFactory.cpp
```diff
@@ -484,6 +487,44 @@ class TransportFactoryImpl final : public TransportFactory,
...
+            {
+                endpoint->unregisterListener(&listener);
+            }
+        }
+    }
+
+    void unregisterIceListener(ServerEndpoint::IEvents& listener, const std::string& ufrag) override
+    {
+        for (auto endpoint : _tcpServerEndpoints)
```
RicardoMDomingues, line 522:
```suggestion
        for (auto& endpoint : _tcpServerEndpoints)
```

transport/ProbeServer.cpp
stepanaksenov: Fixed.

transport/TransportFactory.cpp
stepanaksenov: Fixed.

transport/ProbeServer.cpp
stepanaksenov: I can remove all the mutexes from ProbeServer.

transport/ProbeServer.cpp
stepanaksenov: We can probably add a readiness check in RequestHandler or even expose it to MBR as separate endpoint and check on smb startup. Honestly, I'd prefer to add it in another PR.

bridge/endpointActions/GetProbingInfo.cpp
stepanaksenov: Now we use "Content-type" in all the endpoints. Should I change all of them now?

bridge/endpointActions/GetProbingInfo.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    nlohmann::json candidatesJson = nlohmann::json::array();
+
+    for (auto candidate : probeServer.getProbeCandidates())
+    {
+        nlohmann::json candidateJson;
+        candidateJson["protocol"] = candidate.transport == ice::TransportType::TCP ? "tcp" : "udp";
+        candidateJson["ip"] = candidate.address.ipToString();
+        candidateJson["port"] = candidate.address.getPort();
+        candidatesJson.push_back(candidateJson);
```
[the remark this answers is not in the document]
stepanaksenov: As @olofkallander recommended I left only those properties required for RTT measurement. The "candidates" won't be used for real ICE.

transport/ProbeServer.cpp
stepanaksenov: I will try to add it.

transport/TransportFactory.cpp
stepanaksenov: OK

transport/TransportFactory.cpp
stepanaksenov: OK

transport/TransportFactory.cpp
stepanaksenov: OK

transport/TransportFactory.cpp
stepanaksenov: OK

transport/TransportFactory.cpp
stepanaksenov: OK

transport/TransportFactory.cpp
stepanaksenov: OK

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,224 @@
...
+void ProbeServer::onRegistered(Endpoint& endpoint)
+{
+    const std::lock_guard<std::mutex> locker(_endpointsLock);
+
+    if (endpoint.getTransportType() != ice::TransportType::UDP)
+    {
+        return;
+    }
+
+    auto port = endpoint.getLocalPort().getPort();
```
olofkallander, line 58: This will not work. SMB can open the same port on multiple interfaces.

transport/ProbeServer.cpp
olofkallander, line 76: The publicIpv6 config may be empty. TransportImpl still adds the local IP as a candidate, which in case of IPv6 is the same as the public. 
For ipv4 this scheme may work as the private IPv4 is not much point in listing as candidate when client cannot reach it. However, if SMB would be placed on LAN at customer, the local ipv4 is of interest. Perhaps we should not sut that door. SMB does expose the private IP to the clients joining, so this should do the same out of symetry.

bridge/endpointActions/GetProbingInfo.cpp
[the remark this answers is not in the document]
RicardoMDomingues: Well that's true. But not sure if we have to generate that information anyway to generate the sdp. Then if can provide, it would be easier to generate an sdp. And if you have multiple endpoints with private IPs and public IPs at least the ICE type would be useful. @olofkallander, wdyt?

transport/ProbeServer.cpp
stepanaksenov: So we should suggest both ipv4 and ipv6 local IPs as probe candidates as well?

stepanaksenov: So we should suggest both ipv4 and ipv6 local IPs as probe candidates as well?

bridge/endpointActions/GetProbingInfo.cpp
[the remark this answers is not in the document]
stepanaksenov: Is there a point of having separate probing candidates at all then? We could build list of regular ICE candidates for all the TCP and UDP ports on startup for chosen (generated) ICE credentials; and have TransportFactory to expose them to API.

transport/TransportFactory.cpp
```diff
@@ -484,6 +484,44 @@ class TransportFactoryImpl final : public TransportFactory,
         endpoint->stop(this);
     }
 
+    void registerIceListener(Endpoint::IEvents& listener, const std::string& ufrag) override
+    {
+        for (auto& endpoints : _sharedEndpoints)
+        {
+            for (auto& endpoint : endpoints)
```
olofkallander, line 491: I pondered this a bit. Each vector in _sharedEndpoints contains all endpoints for all interfaces but with a specific port eg. 10000. The next vector will be the same set of interfaces but port 10001 for example. 
There is no point in client trying to reach all interfaces multiple times. It is ok if the probes are all on the first port (10000). The multi-port is to spread load from all RTP. For ICE probes it does not matter.
Once this test probe is done, the client will restart ICE anyway on the port selected for join.
Ie. do not loop over _sharedEndpoints, take the first one if it exists.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,249 @@
...
+}
+
+void ProbeServer::onIceTcpConnect(std::shared_ptr<Endpoint> endpoint,
+    const SocketAddress& source,
+    const SocketAddress& target,
+    memory::UniquePacket packet)
+{
+    if (endpoint->getTransportType() == ice::TransportType::TCP)
+    {
+        replyStunOk(*endpoint, source, std::move(packet));
```
olofkallander, line 151: We talked about the re-register of ufrag in tcp server, right?
I think we concluded that the ufrag must be registered again with the tcp server here since TCPserver thinks this is the final connection for that ufrag and automatically unregistered the ufrag.

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,249 @@
...
+        std::remove_if(_probeCandidates.begin(), _probeCandidates.end(), [port](const transport::ProbeCandidate& c) {
+            return c.transport == ice::TransportType::UDP && c.address.getPort() == port;
+        });
+    }
+}
+
+// ServerEndpoint::IEvents
+void ProbeServer::onServerPortRegistered(ServerEndpoint& endpoint)
+{
+    auto port = endpoint.getLocalPort().getPort();
```
olofkallander, line 107: we should remember this ServerEndpoint so we can re-register ufrag on it later in onTcpConnect.

transport/TransportFactory.cpp
stepanaksenov: Yes, I think we discussed it with @RicardoMDomingues before as initially I only took first UDP endpoint. The concern was that all the probes will be sent to single port. But I can change it back to one port.

RicardoMDomingues: Yes, I mean we have discussed about opening a range of ports in smb because we have already detected our send limit in a single socket is not that high.
Opening different ports on same interface should generate candidates with same foundation. If CS see different candidates with same foundation, it can round robin the candidates provided to client and even provide 2 just to reduce the probability of some glitch caused by an overloaded socket, @olofkallander, wdyt?

bridge/ApiRequestHandler.cpp
stepanaksenov: Changed to "ice-candidates"

transport/ProbeServer.cpp
stepanaksenov: I changed the implementation to pass TCP connection to maintenance thread via queue, as discussed.

transport/ProbeServer.cpp
stepanaksenov: Removed the mutexes.

bridge/endpointActions/GetProbingInfo.cpp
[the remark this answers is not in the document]
stepanaksenov: I changed it to report all the properties we report for ICE candidates in transport description.

transport/ProbeServer.cpp
stepanaksenov: Fixed.

transport/ProbeServer.cpp
stepanaksenov: I removed lists of UDP/TCP ports and instead keep single ICE candidates list (vector).

transport/ProbeServer.cpp
stepanaksenov: Updated to add candidates with local IPs (HOST type).

transport/ProbeServer.cpp
stepanaksenov: Yes, I re-register in `onServerPortUnregistered `callback.

`TcpServerEndpoint `removes listener for the `ufrag` after `onIceTcpConnect ` is called. Adding listener here wouldn't work because of duplicated keys (ufrag) as I understand.

transport/TransportFactory.cpp
stepanaksenov: Changed to pass endpoints for 1st port only.

transport/ice/IceCandidate.h
```diff
@@ -67,6 +67,30 @@ class IceCandidate
     const char* getFoundation() const { return _foundation; }
     void setFoundation(const char* f) { std::strncpy(_foundation, f, MAX_FOUNDATION); }
 
+    static uint32_t computeCandidatePriority(IceCandidate::Type type,
+        int localInterfacePreference,
+        IceComponent component,
+        TransportType transportType)
+    {
+        int typePreference = 0;
```
olofkallander, line 75: Can we move the body to the cpp file?

transport/ProbeServer.cpp
olofkallander: ok

olofkallander:
Do we have any unit test that calls on the HTTP API to get the ice-candidates and check that they are somewhat correct.
A unit test could also create a sort of test client, perhaps reuse SfuClient to test the ICE connectivity with the candidates provided. You have to provide a onfig.json that allows opening of 127.0.0.1, but there are tests doing that.

stepanaksenov:
> Do we have any unit test that calls on the HTTP API to get the ice-candidates and check that they are somewhat correct. A unit test could also create a sort of test client, perhaps reuse SfuClient to test the ICE connectivity with the candidates provided. You have to provide a onfig.json that allows opening of 127.0.0.1, but there are tests doing that.

I've added an integration test. As discussed it queries API for ICE candidates and invokes connect() on transport to trigger ICE probing, then verifies that best RTT is greater than 0.

bridge/endpointActions/GetProbingInfo.cpp
olofkallander: I mistook it for part of json. It is part of HTTP response so that is fine.

api/Parser.cpp
```diff
@@ -517,6 +495,32 @@ ConferenceEndpointExtendedInfo parseEndpointExtendedInfo(const nlohmann::json& d
     return endpoint;
 }
 
+EndpointDescription::Ice parseIce(const nlohmann::json& iceJson)
+{
+    api::EndpointDescription::Ice ice;
+    ice._ufrag = iceJson["ufrag"].get<std::string>();
+    ice._pwd = iceJson["pwd"].get<std::string>();
+
+    for (const auto& candidateJson : iceJson["candidates"])
```
RicardoMDomingues, line 504: Please make it optional like it was before. Candidates are not really required

```suggestion
    for (const auto& candidateJson : optionalJsonArray(iceJson, "candidates"))
```

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,369 @@
...
+            }
+        }
+
+        auto now = utils::Time::getAbsoluteTime();
+
+        for (auto it = _tcpConnections.begin(); it != _tcpConnections.end(); ++it)
+        {
+            if (utils::Time::diffGT(it->timestamp,
+                    now,
+                    _iceConfig.probeConnectionExpirationTimeout * utils::Time::sec)) // TODO use config value
```
RicardoMDomingues, line 311: Plese remove TODO

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    for (auto it = _tcpConnections.begin(); it != _tcpConnections.end(); ++it)
+    {
+        it->endpoint->stop(this);
+        _tcpConnections.erase(it);
+    }
+}
+
+// ICE
```
RicardoMDomingues, line 342: I guess this comment can be removed. Not tell us anything about this method

transport/ProbeServer.cpp
```diff
@@ -0,0 +1,369 @@
...
+const std::pair<std::string, std::string>& ProbeServer::getCredentials() const
+{
+    return _credentials;
+}
+
+const std::vector<ice::IceCandidate> ProbeServer::getCandidates()
```
RicardoMDomingues, line 348: We should consider return a const reference instead and avoid copies of vector each time we need the candidates

RicardoMDomingues: Also can you ensure this is thread safe? I mean we are not making a copy (or returning a reference) when the vector is being changed by other thread?

transport/ProbeServer.cpp
stepanaksenov: removed

transport/ProbeServer.cpp
stepanaksenov: that's for structuring the file

RicardoMDomingues (approved):
LGTM

PR #133, opened 2022-08-16
Barbell2

olofkallander:


bridge/engine/SimulcastStream.h
```diff
@@ -50,6 +50,19 @@ struct SimulcastStream
...
+    {
+        for (size_t i = 0; i < _numLevels; ++i)
+        {
+            if (_levels[i]._ssrc == ssrc)
+            {
+                return i;
+            }
+        }
+
+        return 0;
```
RicardoMDomingues, line 63: This zero sounds error prone. imo it could be an optional or change return type to `int32_t` and return `-1`

RicardoMDomingues (approved):
LGTM

PR #134, opened 2022-08-17
RTC-12973 Share min uplink bitrate estimate among barbells

reddvl1980:
Take barbell's min uplink estimates into account when calculating TMMBR.

### DIFF
--- a/api/DataChannelMessage.h
+++ b/api/DataChannelMessage.h
@@ -132,6 +132,13 @@ inline void makeLastNEnd(utils::StringBuilder<1024>& outMessage)
 #endif
 }
 
+inline void makeMinUplinkBitrate(utils::StringBuilder<1024>& outMessage, const uint32_t maxBitrateKbps)
+{
+    outMessage.append("{\"type\":\"min-uplink-bitrate\", \"bitrateKbps\":\"");
+    outMessage.append(maxBitrateKbps);
+    outMessage.append("\"}");
+}
+
 } // namespace DataChannelMessage
 
 } // namespace api
--- a/api/DataChannelMessageParser.cpp
+++ b/api/DataChannelMessageParser.cpp
@@ -74,6 +74,23 @@ nlohmann::json::const_iterator getEndpointMessagePayload(const nlohmann::json& m
     }
 }
 
+bool isUserMediaMap(const utils::SimpleJson& json)
+{
+    char messageType[64];
+    return json["type"].getString(messageType) && 0 == std::strcmp(messageType, "user-media-map");
+}
+
+bool isMinUplinkBitrate(const utils::SimpleJson& json)
+{
+    char messageType[64];
+    return json["type"].getString(messageType) && 0 == std::strcmp(messageType, "min-uplink-bitrate");
+}
+
+uint32_t getMinUplinkBirate(const utils::SimpleJson& json)
+{
+    return json["bitrateKbps"].getInt(std::numeric_limits<uint32_t>::max());
+}
+
 } // namespace DataChannelMessageParser
 
 } // namespace api
--- a/bridge/engine/EngineBarbell.h
+++ b/bridge/engine/EngineBarbell.h
@@ -68,6 +68,8 @@ struct EngineBarbell
     bridge::RtpMap videoRtpMap;
     bridge::RtpMap videoFeedbackRtpMap;
 
+    uint32_t minEstimatedUplinkBitrate;
+
     static const char* barbellTag;
 };
 
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -1,5 +1,6 @@
 #include "bridge/engine/EngineMixer.h"
 #include "api/DataChannelMessage.h"
+#include "api/DataChannelMessageParser.h"
 #include "bridge/RecordingDescription.h"
 #include "bridge/engine/ActiveMediaList.h"
 #include "bridge/engine/AudioForwarderReceiveJob.h"
@@ -35,6 +36,7 @@
 #include "rtp/RtcpHeader.h"
 #include "rtp/RtpHeader.h"
 #include "transport/Transport.h"
+#include "transport/TransportImpl.h"
 #include "transport/recp/RecControlHeader.h"
 #include "transport/recp/RecDominantSpeakerEventBuilder.h"
 #include "transport/recp/RecStartStopEventBuilder.h"
@@ -172,7 +174,7 @@ EngineMixer::EngineMixer(const std::string& id,
       _engineVideoStreams(maxStreamsPerModality),
       _engineDataStreams(maxStreamsPerModality),
       _engineRecordingStreams(maxRecordingStreams),
-      _engineBarbells(16),
+      _engineBarbells(maxNumBarbells),
       _ssrcInboundContexts(maxSsrcs),
       _audioSsrcToUserIdMap(ActiveMediaList::maxParticipants),
       _localVideoSsrc(localVideoSsrc),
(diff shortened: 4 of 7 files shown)


api/DataChannelMessage.h
```diff
@@ -132,6 +132,13 @@ inline void makeLastNEnd(utils::StringBuilder<1024>& outMessage)
 #endif
 }
 
+inline void makeMinUplinkBitrate(utils::StringBuilder<1024>& outMessage, const uint32_t maxBitrateKbps)
+{
+    outMessage.append("{\"type\":\"min-uplink-bitrate\", \"bitrateKbps\":\"");
```
olofkallander, line 137: Try using R(""). It allows you to drop the \ and format the json in more readable form.

api/DataChannelMessageParser.cpp
```diff
@@ -74,6 +74,23 @@ nlohmann::json::const_iterator getEndpointMessagePayload(const nlohmann::json& m
...
+
+bool isMinUplinkBitrate(const utils::SimpleJson& json)
+{
+    char messageType[64];
+    return json["type"].getString(messageType) && 0 == std::strcmp(messageType, "min-uplink-bitrate");
+}
+
+uint32_t getMinUplinkBirate(const utils::SimpleJson& json)
+{
+    return json["bitrateKbps"].getInt(std::numeric_limits<uint32_t>::max());
```
olofkallander, line 91: It is usually nice to use a num limit like this, but I just worry a bit that it may cause havoc somewhere.
What if we set it to 100Mbps. Seems high and safe enough.

bridge/engine/EngineMixer.cpp
```diff
@@ -1591,11 +1594,17 @@ void EngineMixer::onSctpMessage(transport::RtcTransport* sender,
         }
         auto packet = webrtc::makeUniquePacket(streamId, payloadProtocol, data, length, _sendAllocator);
         auto header = reinterpret_cast<webrtc::SctpStreamMessageHeader*>(packet->get());
-        auto s = reinterpret_cast<char*>(header->data());
-        if (std::strstr(s, "user-media-map") != nullptr)
+        const auto s = reinterpret_cast<char*>(header->data());
+
+        size_t payloadLen = length - (s - reinterpret_cast<const char*>(data));
```
olofkallander, line 1599: const

bridge/engine/EngineMixer.cpp
```diff
@@ -2043,8 +2052,20 @@ void EngineMixer::processBarbellSctp(const uint64_t timestamp)
...
         auto header = reinterpret_cast<webrtc::SctpStreamMessageHeader*>(packetInfo.packet()->get());
-        onBarbellUserMediaMap(packetInfo.transport()->getEndpointIdHash(),
-            reinterpret_cast<const char*>(header->data()));
+        auto message = reinterpret_cast<const char*>(header->data());
+        auto messageJson = utils::SimpleJson::create(message, strlen(message));
+
+        if (api::DataChannelMessageParser::isUserMediaMap(messageJson))
+        {
+            return onBarbellUserMediaMap(packetInfo.transport()->getEndpointIdHash(),
+                reinterpret_cast<const char*>(header->data()));
```
olofkallander, line 2061: just pass message

bridge/engine/EngineMixer.cpp
```diff
@@ -2043,8 +2052,20 @@ void EngineMixer::processBarbellSctp(const uint64_t timestamp)
...
+        }
+
+        if (api::DataChannelMessageParser::isMinUplinkBitrate(messageJson))
+        {
+            return onBarbellMinUplinkEstimate(packetInfo.transport()->getEndpointIdHash(),
+                reinterpret_cast<const char*>(header->data()));
```
olofkallander, line 2067: and message

bridge/engine/EngineBarbell.h
```diff
@@ -68,6 +68,8 @@ struct EngineBarbell
     bridge::RtpMap videoRtpMap;
     bridge::RtpMap videoFeedbackRtpMap;
 
+    uint32_t minEstimatedUplinkBitrate;
```
olofkallander, line 71: The name could mean a variety of things right? We know atm, but perhaps it should be made clear it is not about barbell bandwidth and it is not a bitrate. 
eg.
minClientDownlinkBandwidth

bridge/engine/EngineMixer.cpp
```diff
@@ -3891,4 +3941,23 @@ void EngineMixer::onBarbellUserMediaMap(size_t barbellIdHash, const char* messag
...
+{
+    auto it = _engineBarbells.find(barbellIdHash);
+    if (it == _engineBarbells.end())
+    {
+        return;
+    }
+
+    EngineBarbell& barbell = *(it->second);
+
+    logger::info("received BB msg over barbell %s %zu, json %s",
```
olofkallander, line 3954: this would be quite some spam. debug level perhaps.
And it is specifically minClientDownlinkBandwidth

bridge/engine/EngineMixer.cpp
```diff
@@ -3528,6 +3549,33 @@ void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
...
 }
 
+bool EngineMixer::needToUpdateMinUplinkEstimate(const uint32_t curEstimate, const uint32_t oldEstimate) const
+{
+    // For screensharing (a.k.a. 'slides') we have a minumum allowed bitrate of 900 kbps,
+    // so it does not worth to react on the fluctuation below ~10% of this value.
+    return abs((int64_t)oldEstimate - (int64_t)curEstimate) > 100;
+}
+
+uint32_t EngineMixer::getBarbellsMinUplinkBitrate() const
```
olofkallander, line 3559: perhaps 
getMinRemoteClientDownlinkBandwidth 
though we get it by iterating the barbells recording of such value.

bridge/engine/EngineMixer.cpp
```diff
@@ -3528,6 +3549,33 @@ void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
...
+{
+    auto minBitrate = std::numeric_limits<uint32_t>::max();
+    for (const auto& itBb : _engineBarbells)
+    {
+        minBitrate = std::min(minBitrate, itBb.second->minEstimatedUplinkBitrate);
+    }
+    return minBitrate;
+}
+
+void EngineMixer::reportMinUplinkToBarbells(const uint32_t minUplinkEstimate) const
```
olofkallander, line 3569: and similar name change here

bridge/engine/EngineMixer.cpp
reddvl1980: renamed

bridge/engine/EngineBarbell.h
reddvl1980: Fixed.

api/DataChannelMessage.h
reddvl1980: It works :)

bridge/engine/EngineMixer.cpp
reddvl1980: Changed to debug level. What did yo mean "specifically..."? We'll see JSON in the log.

bridge/engine/EngineMixer.cpp
```diff
@@ -2043,8 +2052,18 @@ void EngineMixer::processBarbellSctp(const uint64_t timestamp)
     for (IncomingPacketInfo packetInfo; _incomingBarbellSctp.pop(packetInfo);)
     {
         auto header = reinterpret_cast<webrtc::SctpStreamMessageHeader*>(packetInfo.packet()->get());
-        onBarbellUserMediaMap(packetInfo.transport()->getEndpointIdHash(),
-            reinterpret_cast<const char*>(header->data()));
+        auto message = reinterpret_cast<const char*>(header->data());
```
RicardoMDomingues, line 2055:
```suggestion
        auto message = header->getMessage();
```

bridge/engine/EngineMixer.cpp
```diff
@@ -2043,8 +2052,18 @@ void EngineMixer::processBarbellSctp(const uint64_t timestamp)
...
+        auto messageJson = utils::SimpleJson::create(message, strlen(message));
```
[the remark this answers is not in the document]
reddvl1980: Yes, I think it worth it. I did include it since I was worried, that since we can't guarantee that incoming json is indeed C-string, something like `utils::SimpleJson::create(const char*)` would tempt it to be used when it is not safe.

PR #141, opened 2022-08-19
RTC-13237 Fix Stream Director

reddvl1980:
Make `StreamDirector::isSsrcUsed` take into account highest active quality, which might be below the desired one.

### DIFF
--- a/bridge/engine/EngineStreamDirector.h
+++ b/bridge/engine/EngineStreamDirector.h
@@ -335,57 +335,41 @@ public:
         const bool isSenderInLastNList,
         const size_t numRecordingStreams)
     {
+        if (_participantStreams.find(senderEndpointIdHash) == _participantStreams.end())
+        {
+            DIRECTOR_LOG("isSsrcUsed, %u false, endpoint removed", "EngineStreamDirector", ssrc);
+            return false;
+        }
+
         const auto quality = getQualityLevel(ssrc);
         const auto highestAvailableQuality = highestActiveQuality(senderEndpointIdHash, ssrc);
+        if (highestAvailableQuality == dropQuality)
+        {
+            DIRECTOR_LOG("isSsrcUsed, %u false, ssrc not found", "EngineStreamDirector", ssrc);
+            return false;
+        }
 
-        // We serve low and mid for unpinned according to the ConfigLadder.
-        if (isUnpinnedQualityUsed(quality, highestAvailableQuality, senderEndpointIdHash, isSenderInLastNList))
+        if (isUsedForUnpinnedVideo(quality, highestAvailableQuality, senderEndpointIdHash, isSenderInLastNList))
         {
             DIRECTOR_LOG("isSsrcUsed, %u default", "EngineStreamDirector", ssrc);
             return true;
         }
 
-        // Pinned endpoint, cold be served in ANY quality, depending on the estimated bandwidth.
-        const auto reversePinMapItr = _reversePinMap.find(senderEndpointIdHash);
-        if (reversePinMapItr != _reversePinMap.end() && reversePinMapItr->second == 0)
+        if (isUsedForPinnedVideo(quality, highestAvailableQuality, senderEndpointIdHash))
         {
-            // This participant used to be pinned, but is no more.
-            DIRECTOR_LOG("isSsrcUsed, %u used to be pinned, but no more", "EngineStreamDirector", ssrc);
-            return false;
+            DIRECTOR_LOG("isSsrcUsed, %u pinned %s",
+                "EngineStreamDirector",
+                ssrc,
+                lowQuality == quality       ? "low"
+                    : midQuality == quality ? "mid"
+                                            : "high");
+            return true;
         }
 
-        // Find who pinned this sender, and which quality does it want.
-        for (const auto& pinMapEntry : _pinMap)
+        if (isUsedForRecordingSlides(ssrc, senderEndpointIdHash, numRecordingStreams))
         {
-            auto const& pinnedBy = pinMapEntry.first;
-            auto const& sender = pinMapEntry.second;
-            if (pinnedBy == senderEndpointIdHash || sender != senderEndpointIdHash)
-            {
-                continue;
-            }
-
-            const auto& participant = _participantStreams.find(pinnedBy);
-            if (participant != _participantStreams.end())
-            {
-                if (std::min(participant->second.desiredHighestEstimatedPinnedLevel, highestAvailableQuality) ==
-                    quality)
-                {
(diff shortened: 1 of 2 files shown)


bridge/engine/EngineStreamDirector.h
```diff
@@ -873,11 +882,14 @@ class EngineStreamDirector
         // Unpinned, belogns to lastN and some of the participants needs this quality.
         for (const auto& participant : _participantStreams)
         {
-            if (participant.first != senderEndpointIdHash && participant.second.desiredUnpinnedLevel == quality)
+            if (participant.first != senderEndpointIdHash &&
```
RicardoMDomingues, line 885: this comment is about line `877` (I can't add a comment on that line because it was not changed on this PR). The code is:
```
        if (anyParticipantsWithoutPinTarget())
        {
            return true;
        }
```

Could you explain to me why we return true if there is a participant without pin target?
It's seems to me if this quality is higher than `desiredUnpinnedLevel` in all participants, it will not be forwarder regardless whether a participant has a pin target or not

bridge/engine/EngineStreamDirector.h
```diff
@@ -856,10 +859,16 @@ class EngineStreamDirector
     /**
      * Checks for ssrcs belonging to a default level if there are participants without pin targets.
      */
-    inline bool isUnpinnedQualityUsed(const uint32_t quality,
+    inline bool isUnpinnedQualityUsed(const size_t quality,
+        const size_t highestAvailableQuality,
         const size_t senderEndpointIdHash,
         const bool isSenderInLastNList)
     {
+        if (quality == highQuality)
```
RicardoMDomingues, line 867:
```suggestion
        if (quality == highestAvailableQuality)
```

If the quality is equal to `highestAvailableQuality` then quality is the pinned qualitty

bridge/engine/EngineStreamDirector.h
```diff
@@ -336,9 +336,10 @@ class EngineStreamDirector
         const size_t numRecordingStreams)
     {
         const auto quality = getQualityLevel(ssrc);
+        const auto highestAvailableQuality = highestActiveQuality(senderEndpointIdHash, ssrc);
 
-        // We server low and mid for unpinned according to the ConfigLadder.
-        if (highQuality != quality && isUnpinnedQualityUsed(quality, senderEndpointIdHash, isSenderInLastNList))
+        // We serve low and mid for unpinned according to the ConfigLadder.
+        if (isUnpinnedQualityUsed(quality, highestAvailableQuality, senderEndpointIdHash, isSenderInLastNList))
```
RicardoMDomingues, line 342: This is my suggestion for readability 
```
cont bool isPinnedQuality = quality == highestAvailableQuality;
if (!isPinnedQuality) {
    return isUnpinnedQualityUsed(quality, highestAvailableQuality, senderEndpointIdHash, isSenderInLastNList);

}

File: bridge/engine/EngineStreamDirector.h [hunk 4]
```diff
@@ -873,11 +882,14 @@ class EngineStreamDirector
         // Unpinned, belogns to lastN and some of the participants needs this quality.
         for (const auto& participant : _participantStreams)
         {
-            if (participant.first != senderEndpointIdHash && participant.second.desiredUnpinnedLevel == quality)
+            if (participant.first != senderEndpointIdHash &&
+                quality == std::min(participant.second.desiredUnpinnedLevel, highestAvailableQuality))
```
RicardoMDomingues, line 886: other check we can add if `senderEndpointIdHash` is pinned for this participant. If it's, the unpinned will not be used for this participant

bridge/engine/EngineStreamDirector.h
reddvl1980: There is a previous check: `if (!isSenderInLastNList)
        {
            return false;
        }`
Together it gives following rule:
_if current sender belongs to LastN set __AND__ there participants w/o pins_ - return `true`. (As if "yes, unpinned quality from this sender is indeed used).

In more details: participants w/o pin will receive _only_ videos from the LastN set, and only in non-high quality.

reddvl1980: Also, this rule: `if (!isSenderInLastNList) { return false; }` immediately filters out senders if they are not in the LastN. Theirs SSRC's still might be used, but as "pinned" and there is a separate check for this in the 'isSsrcUsed'.

bridge/engine/EngineStreamDirector.h
reddvl1980: `if (quality == highestAvailableQuality)` this is not right, as it will result in some medium and low streams considered "unused for unpinned".  The change you propose basically says: _"we never stream highest available quality to unpinned"_, which is not true. We stream say, _low_, to unpinned, and highest available quality might be low as well (as it is indeed for the screen sharing).

bridge/engine/EngineStreamDirector.h
reddvl1980: See previous comment.
I'm going to refactor the code to make it more readable and intentions clear. I see that now it's quite confusing,

bridge/engine/EngineStreamDirector.h
reddvl1980: Your logic is correct, but as you said: _".. If it's, the unpinned will not be used for THIS participant"_, we can not early break on it, because we need to know whether this quality is needed at all (by at least one participant). So the fact that one participant needs _pinned_ quality from our sender, does not exclude possibility that some other participants need _unpinned_ quality from this very sender.

bridge/engine/EngineStreamDirector.h
RicardoMDomingues: Yes, but if all participants are interested in lowQality, this will return `true` for midQuality. or if all want midQaulity, the lowQuality is not necessary.

bridge/engine/EngineStreamDirector.h
RicardoMDomingues: The name creates some confusion. Is more about if non high quality is used instead of "unpinned" quality

bridge/engine/EngineStreamDirector.h
RicardoMDomingues: Yes, I know I am just saying could continue iterate because imagine all participants are able to receive the highest quality for dominant speaker, then packets for this layer does not need to be decrypted.

bridge/engine/EngineStreamDirector.h
reddvl1980: Fixed.

olofkallander (approved):

RicardoMDomingues (approved):

PR #143, opened 2022-08-22
barbell PLI

olofkallander:
RTC-12934

### DIFF
--- a/bridge/engine/SsrcOutboundContext.h
+++ b/bridge/engine/SsrcOutboundContext.h
@@ -86,6 +86,25 @@ public:
     uint64_t lastSendTime;
     bool markedForDeletion;
     bool idle;
+
+    struct PliMetric
+    {
+        uint64_t userRequestTimestamp = 0;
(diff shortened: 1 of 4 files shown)


bridge/engine/SsrcOutboundContext.h
```diff
@@ -86,6 +86,25 @@ class SsrcOutboundContext
...
+
+    struct PliMetric
+    {
+        uint64_t userRequestTimestamp = 0;
+        uint64_t triggerTimestamp = 0;
+        uint64_t keyFrameTimestamp = 0;
+
+        uint64_t getDelay(const uint64_t timestamp) const
+        {
+            if (userRequestTimestamp != 0 && static_cast<int64_t>(userRequestTimestamp - keyFrameTimestamp) > 0)
```
reddvl1980, line 98: What if have both? (triggered and user-requested?)

olofkallander: yes, then we log the user requested one

PR #146, opened 2022-08-23
Remove sleep waits

RicardoMDomingues:
Main changes:

- Remove sleep waits that blocks mixerManager thread
- Implement `jobmanager::AsyncWaiter`
- Do not send more than on Mixer timeout message

**WIP:** Unit tests


bridge/MixerManager.cpp
```diff
@@ -354,25 +373,28 @@ void MixerManager::engineMessageMixerRemoved(const EngineMessage::Message& messa
...
         std::string mixerId(command.mixer->getId()); // copy id string to have it after EngineMixer is deleted
-        auto findResult = _mixers.find(mixerId);
-        if (findResult != _mixers.end())
+        auto mixer = eraseAndGet(_mixers, mixerId);
+        if (mixer)
         {
-            mixer = std::move(findResult->second);
             mixer->stopTransports(); // this will stop new packets from coming in
-            if (!mixer->waitForAllPendingJobs(1000))
+            if (mixer->hasPendingJobs())
```
reddvl1980, line 380: Consider renaming to 'pop' (though it's usually a queue method, it looks like it would fit here perfectly).

RicardoMDomingues: do you mean on line `376` to replace `eraseAndGet`?

bridge/Mixer.cpp
```diff
@@ -2167,25 +2162,33 @@ void Mixer::engineRecordingStreamRemoved(EngineRecordingStream* engineStream)
...
+        if (transportEntry.second->hasPendingJobs())
         {
-            logger::error("RecordingStream id %s did not finish pending jobs in time. count=%u. Continuing "
-                          "deletion anyway.",
-                _loggableId.c_str(),
-                stream->_id.c_str(),
-                transportEntry.second->getJobCounter().load());
+            initializeOrIncrement(pendingTransportCounter, stream->_endpointIdHash);
+            auto asyncWaitTask = PendingJobsAsyncWaitTask(_loggableId, std::move(transportEntry.second), engineRecordingStream, engineStream->id);
+            asyncWaitTask.setOnTaskEndHandler([this, pendingTransportCounter] () {
```
olofkallander, line 2175: It is a bit unintelligable what is happening when you read this code. Maybe we should ponder a bit if we have made it as simple as it can be done.

RicardoMDomingues: I can create a a task specific for recording. I avoided that solution because I was trying having the asynWaiter with a template task so I couldn't have a different task type for recording use case.
I changed that and now it can receive any task that extends of AsyncWaitTask so I will create a specialisation for recording and will simplify this

jobmanager/AsyncWaiter.h
```diff
@@ -0,0 +1,195 @@
...
+}
+
+inline void AsyncWaiter::runCheck()
+{
+    const auto timestamp = utils::Time::getAbsoluteTime();
+    auto itEnd = _tasks.end();
+    auto nextTimeout = _maxCheckInterval;
+    for (auto it = _tasks.begin(); it != itEnd; )
+    {
+        const bool isCompleted = checkTaskCompletion(*it->task);
```
olofkallander, line 122: hasEnded, hasCompleted, or
just 
completed, ended

jobmanager/AsyncWaiter.h
```diff
@@ -0,0 +1,195 @@
...
+            {
+                it->task->onTimeout();
+            }
+
+            *it = std::move(*(itEnd - 1));
+            --itEnd;
+        }
+        else
+        {
+            nextTimeout = std::min(nextTimeout, it->endTime - timestamp);
```
olofkallander, line 136: You will miss setting next check 0ms if the timestamp passed endtime. endtime - timestamp will become huge and we will wait _maxCheckInterval, unless there is another task with slightly less time left.

jobmanager/AsyncWaiter.h
```diff
@@ -0,0 +1,195 @@
...
+
+private:
+    void runCheck();
+    void scheduleNexCheck(uint64_t timeoutTs);
+    void addNewTaskInternal(uint64_t timeout, std::unique_ptr<AsyncWaitTask>&& task);
+    bool checkTaskCompletion(AsyncWaitTask& task);
+
+private:
+    JobQueue _jobQueue;
+    std::vector<AsyncEntry> _tasks;
```
olofkallander, line 99: I am a bit sceptical to this data type. Workerthreads will have to do allocation. Could it be a more static array like StackArray. I know it is not ready for emplace and std:move, but perhaps it is an alternative.

bridge/Mixer.cpp
```diff
@@ -230,69 +260,39 @@ void Mixer::stopTransports()
...
-        }
+    static const auto recordingHasPendingJobsPredicate = [] (auto& recordingStream)
+    {
+            const auto& transports = recordingStream.second->_transports;
+            return std::find_if(transports.begin(), transports.end(), [] (auto& transport) { return transport.second->hasPendingJobs(); }) != transports.end();
+    };
 
-        _engineMixer.flush();
-    }
+    bool hasPendingJobs = _pendingJobsAsyncWaiter.hasTaskWaiting();
```
olofkallander, line 287: Collapsing the whole implementatioin, we could have put all the shared_ptr<RtcTransports> in a vector and use a find_if like below but with one predicate and run this on a timerjob to check it and then set a flag or report to mixermanager when it is done.

jobmanager/AsyncWaiter.h
RicardoMDomingues: Here `it->endTime` is alway bigger than `timestamp`, otherwise the task ended by timeout

jobmanager/AsyncWaiter.h
RicardoMDomingues: I thought a lot about this but I went with a simpler approach  because as is not used by EngineMixer and it creates a std::unique_ptr which needs a dynamic allocation anyway (although the allocation is on callee thread) the deallocation of unique_ptr will be on wt.

if we want to avoid dynamic allocations and have only a pure virtual type (AsyncWaitTask) we perhaps need to use memory pools here as well

bridge/engine/UntypedEngineObject.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace bridge
+{
+
+
+struct UntypedEngineObject
```
olofkallander, line 7: I do not like the term untyped. Why not 
GenericEngineObject

bridge/PendingJobsAsyncWaitTask.h
```diff
@@ -0,0 +1,43 @@
...
+{
+
+class PendingJobsAsyncWaitTask : public jobmanager::AsyncWaitTask
+{
+public:
+    using OnTaskEndHandler = std::function<void()>;
+
+    PendingJobsAsyncWaitTask(const logger::LoggableId& loggableId,
+        const std::shared_ptr<const transport::Transport>& _transport,
+        const std::shared_ptr<UntypedEngineObject>& streamToHold,
```
olofkallander, line 26: Was it imposible to make this a template to avoid poluting the objects we want to store here. It inherits AsyncWaitTask so it still fits into the AsyncWaiter.
You usually pass lambda into task handler, and we do mainly use it for getting destructor called when all references are gone. Am I missing something?

bridge/engine/UntypedEngineObject.h
olofkallander: I think this class should be removed completely and use template PendingJobsAsyncWaitTask

jobmanager/AsyncWaiter.h
```diff
@@ -0,0 +1,195 @@
...
+
+inline void AsyncWaiter::runCheck()
+{
+    const auto timestamp = utils::Time::getAbsoluteTime();
+    auto itEnd = _tasks.end();
+    auto nextTimeout = _maxCheckInterval;
+    for (auto it = _tasks.begin(); it != itEnd; )
+    {
+        const bool isCompleted = checkTaskCompletion(*it->task);
+        const bool isEnded = isCompleted || timestamp >= it->endTime;
```
olofkallander, line 123: we should always account for wrapping of timestamp. Use utils::Time::diffGE

jobmanager/AsyncWaiter.h
olofkallander: We could at least use reserve to have it ready for most likely scenarios.

bridge/Mixer.cpp
```diff
@@ -312,7 +312,7 @@ bool Mixer::addBundleTransportIfNeeded(const std::string& endpointId, const ice:
         return false;
     }
```
olofkallander, line 314: This has to be PRIu64 instead of llu.
llu always cause incompatibility on some platform.

utils/Time.h
```diff
@@ -66,6 +66,26 @@ constexpr bool diffGE(const uint64_t a, const uint64_t b, const uint64_t value)
     return static_cast<int64_t>(b - a) >= static_cast<int64_t>(value);
 }
 
+constexpr bool lt(const uint64_t a, const uint64_t b)
```
RicardoMDomingues, line 69: Add methods to compare two timestamps

jobmanager/AsyncWaiter.h
RicardoMDomingues: I reserve 16 by default. I don't expect having much pending task in simultaneity. Each mixer has their own queue

utils/Time.h
olofkallander: Not that it was correct before, but all of these could and should be based on the diff method at line 44.
including diffLT, diffLE.

jobmanager/JobManager.h
```diff
@@ -114,6 +114,18 @@ class JobManager
         return nullptr;
     }
 
+    Job* tryFetchNoWait()
+    {
+        bool isRunning = _running.load(std::memory_order::memory_order_relaxed);
```
olofkallander, line 119: const

jobmanager/JobQueue.h
```diff
@@ -59,7 +60,30 @@ class JobQueue
...
+        // Wait 1ms to this semaphore to be released, if it's not
+        // released after that we will check if this a worker thread
+        // and yield it to process other jobs that jobManager may have.
+        // If it is not an worker thread then will wait forever until semaphore is released
+        if (!sema.wait(1))
+        {
+            if (WorkerThread::isWorkerThread())
+            {
+                bool waiting = true;
+                while (waiting)
```
olofkallander, line 73: I prefer 
for(bool waiting = true;wating;) 
to keep the variable inside the block

jobmanager/WorkerThread.cpp
```diff
@@ -45,6 +53,30 @@ void WorkerThread::run()
     {
         logger::error("unknown exception", "WorkerThread");
     }
+
+    workerThreadHandler = nullptr;
+}
+
+bool WorkerThread::yield()
```
olofkallander, line 60: put a comment here that it returns true if a job was processed. That will help when reading code at other places, without having excessive comments what is happening. The IDE will show me what the return value of yield means.

jobmanager/JobQueue.h
```diff
@@ -59,7 +60,30 @@ class JobQueue
...
+                {
+                    // Let's yield thread until StopJob is done.
+                    // check semaphore without wait in the case a job has been
+                    // processed, otherwise wait for 1ms until next yield
+                    const bool switchedJobExecution = WorkerThread::yield();
```
olofkallander, line 78: switchJobExecution means nothing.
jobProcessed =

jobmanager/JobQueue.h
```diff
@@ -59,7 +60,30 @@ class JobQueue
     {
         concurrency::Semaphore sema;
         addJob<StopJob>(sema, _running);
-        sema.wait();
+
+        // Wait 1ms to this semaphore to be released, if it's not
+        // released after that we will check if this a worker thread
+        // and yield it to process other jobs that jobManager may have.
+        // If it is not an worker thread then will wait forever until semaphore is released
+        if (!sema.wait(1))
```
olofkallander, line 68: why do the wait of 1ms at all if we are workerthread? If this is a workerthread we can just as well go on with the processing of jobs. No one should try add jobs to this queue after ~JobQueue has started to execute.

jobmanager/WorkerThread.cpp
olofkallander: Otherwise, brilliant.

bridge/PendingJobsAsyncWaitTask.h
```diff
@@ -0,0 +1,110 @@
...
+template <class T, class... Args>
+bool PendingJobsAsyncWaitTask<T, Args...>::canComplete()
+{
+    return !_pendingJobsChecker->hasPendingJobs();
+}
+
+template <class T, class... Args>
+void PendingJobsAsyncWaitTask<T, Args...>::onComplete()
+{
+    logger::info("PendingJobsAsyncWaitTask %u has completed", _loggableId, _taskId);
```
olofkallander, line 65: I do not think we can have this on info level in general. Even debug risk spamming a lot depending on how we use this. Perhaps the logic of the specific wait task has to log depending on scenario.

bridge/PendingJobsAsyncWaitTask.h
```diff
@@ -0,0 +1,110 @@
...
+}
+
+template <class T, class... Args>
+void PendingJobsAsyncWaitTask<T, Args...>::onTimeout()
+{
+    logger::error("PendingJobsAsyncWaitTask %u has timed out", _loggableId, _taskId);
```
olofkallander, line 71: It could be, but is not necessarily an error.

bridge/PendingJobsAsyncWaitTask.h
```diff
@@ -0,0 +1,110 @@
...
+      _taskId(_taskId),
+      _pendingJobsChecker(std::move(pendingJobsChecker)),
+      _memoryToHold(std::forward<UTypes>(memoryToHold)...)
+{
+}
+
+template <class T, class... Args>
+bool PendingJobsAsyncWaitTask<T, Args...>::canComplete()
+{
+    return !_pendingJobsChecker->hasPendingJobs();
```
olofkallander, line 59: I ponder a better name for this. 
pendingJobsMonitor 
maybe

bridge/MixerManager.h
```diff
@@ -103,7 +104,11 @@ class MixerManager : public EngineMessageListener
     memory::PacketPoolAllocator& _mainAllocator;
     memory::PacketPoolAllocator& _sendAllocator;
     memory::AudioPacketPoolAllocator& _audioAllocator;
+    jobmanager::AsyncWaiter _asyncWaiter;
+    uint32_t _asyncWaiterTaskCount;
 
+    friend struct MixerDeleteTask;
+    void onMixerReadyToDeletion(std::unique_ptr<Mixer>&& mixer);
```
olofkallander, line 111: onMixerReadyForDeletion

bridge/PendingJobsAsyncWaitTask.h
RicardoMDomingues: I have implemented a log policy in this commit https://github.com/finos/SymphonyMediaBridge/pull/146/commits/8f7d84718b6a991934d134a916962e30e3bc6e86. Not sure if I am happy with it though.

utils/Time.h
RicardoMDomingues: I changed the `diff...` to use `diff`. This ones uses the `diff...`

jobmanager/WorkerThread.cpp
RicardoMDomingues: I added the comment on .h file

jobmanager/JobQueue.h
RicardoMDomingues: Yeah, I thought the same when I was implementing but I decided to give a chance to execute the StopJob with a small wait. But I changes to not wait if there are jobs to run like you suggested

jobmanager/WorkerThread.h
```diff
@@ -15,6 +15,20 @@ class WorkerThread
 
     static double getWaitTime(); // ms
     static double getWorkTime(); // ms
+
+    /**
+     * @brief Checks whether current thread is a WorkerThread or not
+     * @return true if current thread is a worker thread; otherwise, false.
+     */
+    static bool isWorkerThread();
```
olofkallander, line 23: The method name in this case is excellent.
It already says everything that is in the comment.

jobmanager/WorkerThread.h
```diff
@@ -15,6 +15,20 @@ class WorkerThread
...
+
+    /**
+     * @brief Causes the calling thread to yield execution to another job that is ready to run on
+     * jobManager attached to current WorkerThread.
+     * @return true if has executed to another job; otherwise, false.
+     */
+    static bool yield();
```
olofkallander, line 30: Here on the other hand, it is a bit mysterious what the return value means.

PR #148, opened 2022-08-24
Use FakeUdpEndpoint, NetworkLink and FakeNetwork for integration tests.

reddvl1980:


bridge/Bridge.cpp
```diff
@@ -91,6 +92,11 @@ Bridge::~Bridge()
 }
 
 void Bridge::initialize()
+{
+    initialize(std::shared_ptr<transport::EndpointFactory>(new transport::EndpointFactoryImpl()));
```
RicardoMDomingues, line 96: We should use allows `make_shared`. It's more efficient as it use a single heap allocation to `EndpointFactoryImp` and ref counter controller, keep that memory close to each other and in C++14 `std:shared` + `new` as an argument of a function is not exception safe and it can create memory leaks if an exception is thrown.


```suggestion
    initialize(std::make_shared<EndpointFactoryImp>());
```

transport/IUdpEndpoint.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+#include "transport/Endpoint.h"
+#include "transport/RtcSocket.h"
+#include "transport/RtcePoll.h"
+
+namespace transport
+{
+class IUdpEndpoint : public Endpoint, public RtcePoll::IEventListener
```
RicardoMDomingues, line 8: Usually we don't use prefix "I" in this project. I guess it should be called UdpEndpoint and our current `UdpEndpoint` should be renamed to `UdpEndpointImpl` to be consistent

transport/EndpointFactoryImpl.cpp
```diff
@@ -0,0 +1,26 @@
+#include "transport/EndpointFactoryImpl.h"
+#include <memory>
+
+namespace transport
+{
+IUdpEndpoint* EndpointFactoryImpl::createUdpEndpointStatic(jobmanager::JobManager& jobManager,
```
RicardoMDomingues, line 6: Personally I would prefer factories to return unique_ptr instead of raw pointers. Even if you need to store a shared_ptr later, you can move ownership of unique_ptr into a shared_ptr

transport/UdpEndpoint.h
```diff
@@ -3,19 +3,28 @@
...
 {
+class EndpointFactoryImpl;
+
 // end point that can be shared by multiple transports and can route incoming traffic
 class UdpEndpoint : public BaseUdpEndpoint
 {
-public:
+    friend class EndpointFactoryImpl;
+    friend class emulator::FakeEndpointFactory;
     UdpEndpoint(jobmanager::JobManager& jobManager,
```
RicardoMDomingues, line 20: I don't see why the UdpEndpoint needs to have a private constructor and some `friends`. IMHO an object should not "baby-sitting" the other objects to prevent them from mistakes that are not of the responsibility of this object.

Also I didn't find the usage of `UdpEndpoint` on `FakeEndpointFactory`.

RicardoMDomingues (commented):
LGTM in general, only some minor comments

bridge/Bridge.cpp
reddvl1980: Agree, fixed.

transport/IUdpEndpoint.h
reddvl1980: +1. Done

transport/EndpointFactoryImpl.cpp
reddvl1980: The real reason behind returning raw pointer is that we're using custom deleter in the TransportFactory. I found no way how to "append" a custom deleter to existing shared pointer.  And extracting raw from shared only to put it back to another shared, but this time with a custom deleter looks wrong.

transport/UdpEndpoint.h
reddvl1980: The latter is correct, there is no use of `UdpEndpoint` from the `FakeEndpointFactory` -  it's remnant of the development, and I removed it.

Regarding making constructor private - that's the only way to enforce use of the factory, which, in-turn, is the only way to guarantee that all endpoints are created of the same type (either real, or fake).

RicardoMDomingues (approved):

PR #151, opened 2022-08-25
RTC-13244: fix probe when rtx not available

RicardoMDomingues:


bridge/RtpMap.h
```diff
@@ -66,6 +66,8 @@ struct RtpMap
 
     RtpMap(const RtpMap& rtpMap) = default;
 
+    bool isEmpty() const { return format == Format::EMPTY; }
+
     Format format;
     uint16_t payloadType;
```
olofkallander, line 72: I wonder if this should be uint8_t. Or do we expect payload types outside that range? Even payload types > 127 are not valid.

test/integration/emulator/SfuClient.h
```diff
@@ -105,18 +105,55 @@ class SfuClient : public transport::DataReceiver
 
         _remoteVideoSsrc = _channel.getOfferedVideoSsrcs();
         _remoteVideoStreams = _channel.getOfferedVideoStreams();
+        uint32_t localSsrc = _channel.getOfferedLocalSsrc();
+        utils::Optional<uint32_t> slidesSsrc = _channel.getOfferedScreensharingSsrc();
 
         bridge::RtpMap videoRtpMap(bridge::RtpMap::Format::VP8);
         bridge::RtpMap feedbackRtpMap(bridge::RtpMap::Format::VP8RTX);
 
+        bool isLocalSsrcFond = false;
```
olofkallander, line 114: isLocalSsrcFound

bridge/RtpMap.h
RicardoMDomingues: I changed to uin8_t

olofkallander (approved):

PR #155, opened 2022-08-26
Olofk/TimeSource

olofkallander:
time emulator allowing to run tests faster


test/integration/emulator/TimeTurner.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        advance();
+        do
+        {
+            utils::Time::rawNanoSleep(10);
+        } while (_sleepersCount.load() < threadCount);
+    }
+}
+
+void TimeTurner::advance()
```
olofkallander, line 67: I will change this to wait for a sempahore until threads have completed. It allows for faster progress.

reddvl1980: +1

test/integration/emulator/TimeTurner.cpp
```diff
@@ -0,0 +1,114 @@
...
+            _sleeperCount.decrement();
+            slot.semaphore.post();
+        }
+    }
+}
+
+void TimeTurner::shutdown()
+{
+    _running = false;
+    for (auto& slot : _sleepers)
```
olofkallander, line 104: We are releasing all threads now and they will not wait anymore. We do not have many thras that are waiting for specific timestamps before they continue, but if someone adds that, that thread will not complete its loop.

utils/Time.h
```diff
@@ -1,24 +1,48 @@
...
+    virtual std::chrono::system_clock::time_point wallClock() = 0;
+};
+
 namespace Time
 {
 
 void initialize();
-
+void initialize(TimeSource* timeSource);
+void cleanup();
```
olofkallander, line 29: I think I will remove this cleanup and let the time source be owned outside.

test/integration/emulator/TimeTurner.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * You can run this initially if you know how many threads minimum will join the sleep.
+ * It can only be run once.
+ */
+void TimeTurner::waitForThreadsToSleep(uint32_t expectedCount, uint64_t timeoutNs)
+{
+    for (uint32_t i = 0; i < expectedCount - 1; ++i)
+    {
+        _sleeperCount.decrement();
```
RicardoMDomingues, line 51: Now I guess I understand the changes on semaphore, although I think this logic not be archived by a weird implementation of a semaphore. My suggestion is implement that logic here or other type that manage int value and a method that can wait for a specific condition on that value

concurrency/Semaphore.cpp
```diff
@@ -33,31 +30,32 @@ bool Semaphore::wait(uint32_t timeoutMs)
...
+    _count = count;
 }
 
 void Semaphore::post()
 {
     std::lock_guard<std::mutex> locker(_lock);
-    ++_count;
-    _conditionVariable.notify_one();
+    const auto count = _count.fetch_add(1);
+    if (count >= 0)
```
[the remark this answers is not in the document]
olofkallander: When the return value from fetch_add is = 0, it means it is now > 0.
But it should not be atomic in the first place

test/integration/emulator/TimeTurner.cpp
olofkallander: I will create another one. It is just that a sempahore is logically a resource counter. When there are resources available, consumers can take one and continue until the resources are depleted and you will have to wait for resources to become available again.
The semaphore I am after is the same except that it starts with a debt of resources, instead of 0 resources.

test/integration/emulator/TimeTurner.cpp
```diff
@@ -0,0 +1,143 @@
...
+namespace emulator
+{
+
+TimeTurner::TimeTurner() : _timestamp(100), _startTime(std::chrono::system_clock::now()), _running(true) {}
+
+void TimeTurner::nanoSleep(const uint64_t nanoSeconds)
+{
+    if (!_running)
+    {
+        std::this_thread::yield();
```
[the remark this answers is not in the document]
olofkallander: We can let it run but we probably have threads at the end of a test. The thread did expect to pasue anyway so if there is nothing else to do it will continue.

RicardoMDomingues (approved):

reddvl1980 (approved):

PR #157, opened 2022-08-30
fix BB message

olofkallander:


reddvl1980 (approved):

RicardoMDomingues (approved):

transport/sctp/SctpAssociationImpl.cpp
```diff
@@ -371,7 +371,11 @@ SctpAssociationImpl::SctpAssociationImpl(size_t logId,
         _local.inboundStreamCount = cookie.inboundStreams;
         _local.outboundStreamCount = cookie.outboundStreams;
     }
+}
 
+// must only be called once and immediately after creation due to COOKIE ECHO received
+void SctpAssociationImpl::onCookieEcho(const SctpPacket& cookieEcho, const uint64_t timestamp)
```
olofkallander, line 377: This was the main problem. Processing the echo caused callbacks onto IEvents before the object had been returned and still in constructor.

bridge/engine/EngineMixer.cpp
```diff
@@ -1527,12 +1527,17 @@ bool EngineMixer::onSctpConnectionRequest(transport::RtcTransport* sender, uint1
...
 void EngineMixer::onSctpEstablished(transport::RtcTransport* sender)
 {
-    logger::debug("SCTP association established", sender->getLoggableId().c_str());
+    logger::info("SCTP association established %s", _loggableId.c_str(), sender->getLoggableId().c_str());
 
     auto barbellIt = _engineBarbells.find(sender->getEndpointIdHash());
-    if (barbellIt != _engineBarbells.cend() && sender->isDtlsClient())
+    if (barbellIt != _engineBarbells.cend())
     {
+        logger::debug("opening barbell webrtc data channel on %s",
```
olofkallander, line 1535: We have to open webrtc data channel from obth sides simultaneously as we want to send the BB msg right away. 
Option would be to wait for the webrtc data channel to be established from the DTLS client side.

bridge/engine/EngineMixer.cpp
```diff
@@ -1543,26 +1534,11 @@ void EngineMixer::onSctpMessage(transport::RtcTransport* sender,
...
-    // TODO  parse this with wonderful json parser and figure out if this is a barbell UMM
-    // if it is we should handle it locally in onBarbellUserMediaMap
-    if (_engineBarbells.contains(sender->getEndpointIdHash()))
+    if (EngineBarbell::isFromBarbell(sender->getTag()))
     {
-        if (length == 0)
-        {
-            return;
-        }
         auto packet = webrtc::makeUniquePacket(streamId, payloadProtocol, data, length, _sendAllocator);
```
olofkallander, line 1539: send all barbell data stream message directly to EngineMixer. Do not pass them through MixerManager.

olofkallander:
- Make sure sctpAssociation is returned to Transport before processing the cookieEcho as it will cause callbacks into Transports. The latter must see the sctpAssociation.
- Trigger data channel setup from one side when SCTP is established.
- Send BB UMM when data channel setup is complete to avoid sending on different streams, etc
- Process all data channel for barbell directly in EngineMixer without passing through MixerManager

PR #158, opened 2022-08-31
Olofk/ssrclogs

olofkallander:
identifying out of order reception and transmission

### DIFF
--- a/transport/dtls/SrtpClient.cpp
+++ b/transport/dtls/SrtpClient.cpp
@@ -51,6 +51,8 @@ SrtpClient::SrtpClient(SslDtls& sslDtls, IEvents* eventListener)
       _localSrtp(nullptr),
       _nullCipher(true),
       _eventSink(eventListener),
+      _rtpAntiSpam(10, 100),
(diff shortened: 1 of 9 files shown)


transport/dtls/SrtpClient.cpp
```diff
@@ -306,6 +306,11 @@ bool SrtpClient::unprotect(memory::Packet& packet)
             return false;
         }
     }
+    else
+    {
+        return false;
```
RicardoMDomingues, line 311: I guess the only reason to code hit this line is due a bug, right? As we are explicitly handle this case it wouldn't be better write na error log to not be blind about this when it happens?

RicardoMDomingues (approved):

transport/dtls/SrtpClient.cpp
olofkallander: or perhaps assert(false). We will have checked already that it is an rtp packet. 
But there is also a problem if client sneds us real rtp header but rest of packet cannot be unprotected due to it is corrupt. It will cause log spam.

reddvl1980 (approved):

PR #160, opened 2022-08-31
Make IntegrationTest use TimeTurner

reddvl1980:
Scaffolding 'plain' test for TimeTurner use


test/integration/IntegrationTest.cpp
```diff
@@ -65,23 +65,46 @@ IntegrationTest::IntegrationTest()
...
 }
 
 void IntegrationTest::SetUp()
 {
+
+    if (__has_feature(address_sanitizer) || __has_feature(thread_sanitizer))
+    {
+        GTEST_SKIP();
+    }
+#if !ENABLE_LEGACY_API
```
olofkallander, line 80: Is this something we do now only? I mean the tests contain both legacy and SMB API.

test/integration/IntegrationTest.cpp
```diff
@@ -259,824 +292,843 @@ bool isActiveTalker(const std::vector<api::ConferenceEndpoint>& endpoints, const
...
+void IntegrationTest::finalizeSimulation()
+{
+    _internet->stop();
+    utils::Time::initialize();
+    _timeSource.shutdown();
+}
+
+TEST_F(IntegrationTest, plain)
+{
+    runTestInThread(_numWorkerThreads + 4, [this]() {
```
olofkallander, line 329: We must look for a way around this.

olofkallander (approved):

test/integration/IntegrationTest.cpp
reddvl1980: This macros was present in all the tests, basically disabling tests if the build was not done with the _legaci API_ support. So this changes nothing. All tests would be run, but only on the build configs that:
a) does not use thread/address sanitizers
b) supports _legacy API_.

In future, when the _legacy API_ will be deprecated, we'll remove certain tests, refactor others and remove this check at all.

test/integration/IntegrationTest.cpp
reddvl1980: Please be more specific. Are you talking about number of threads we're waiting for or about lambda?

PR #161, opened 2022-08-31
asyncjobs

olofkallander:


jobmanager/Job.h
```diff
@@ -6,13 +6,30 @@
...
+public:
+    JobBase() = default;
+    virtual ~JobBase() {}
+
+    // return true if job needs to be run again
+    virtual bool runOnce() = 0;
+};
+
+class Job : public JobBase
 {
```
olofkallander, line 20: These two classes whould swap names, but I did not want to cause all the refactoring atm.

jobmanager/WorkerThread.h
```diff
@@ -10,18 +12,28 @@ class WorkerThread
...
 private:
     std::atomic<bool> _running;
     jobmanager::JobManager& _jobManager;
 
     void run();
+    bool processJobs();
     static void threadEntry(WorkerThread* instance);
+    uint32_t processYieldedJobs();
 
+    std::array<JobBase*, 512> _yieldedJobs;
```
olofkallander, line 34: question how to dimension this. It could be by config and dynamically allocated as it is done at startup.
With multiple threads there can be a lot of yielded tasks.

jobmanager/WorkerThread.cpp
```diff
@@ -45,6 +95,68 @@ void WorkerThread::run()
...
+// return true if any new jobs were processed
+bool WorkerThread::processJobs()
+{
+    if (_backgroundJobCount == _backgroundJobs.size())
+    {
+        _backgroundJobCount = processBackgroundJobs();
+        if (_backgroundJobCount == _backgroundJobs.size())
+        {
+            // do not take on more jobs as we cannot store them if in case they are be multi-step job
+            return false;
```
RicardoMDomingues, line 110: I would prefer having a vector with 512 preallocated than this fix value. WT already allocates memory in other places. To me it should preallocate a value that will work in most of times in a healthy system but allow it to grow until a bigger limit in case if we experience a spike.

Also it does not make sense for me avoid allocate memory to avoid the time of a syscall can take in exchange of delaying the job processing (which probably will cause a much bigger delay than the allocation itself)

jobmanager/WorkerThread.cpp
```diff
@@ -45,6 +95,68 @@ void WorkerThread::run()
...
+        bool runAgain = job->runStep();
+        if (!runAgain)
+        {
+            _jobManager.freeJob(job);
+        }
+        else
+        {
+            for (uint32_t i = 0; i < _backgroundJobs.size(); ++i)
+            {
+                if (!_backgroundJobs[i])
```
RicardoMDomingues, line 132: Leaking jobs here. For example if _backgroundJobs only have space for 1 job and this cycle of 10 jobs need to add more than 1, the others will be leaked.

jobmanager/Job.h
```diff
@@ -1,18 +1,38 @@
...
+// Use this to implement long running jobs that can continue where it left off.
+// Track the state inside your job to execute the next step when run again.
+class MultiStepJob
+{
+public:
+    MultiStepJob() = default;
+    virtual ~MultiStepJob() {}
+
+    // return true if job needs to be run again
+    virtual bool runStep() = 0;
```
reddvl1980, line 18: Good class & method name!

test/jobmanager/JobManagerTest.cpp
```diff
@@ -279,3 +279,112 @@ TEST_F(JobManagerTest, timers)
     end = utils::Time::getAbsoluteTime();
     EXPECT_NEAR(end - start, timeout * 3, 30 * ms);
 }
+
+namespace
+{
+class YieldJob : public MultiStepJob
+{
+public:
+    YieldJob(std::atomic_int& counter) : _start(utils::Time::getAbsoluteTime()), _counter(counter) { ++counter; }
```
reddvl1980, line 288: Consider uisng "_counter(++counter)"

reddvl1980 (approved):
LGTM

jobmanager/WorkerThread.cpp
olofkallander: Suspending the job by putting it in the _backgroundJobs will actually not cost anything for the WorkerThread. Unless, it has to allocate memory.
The memory allocations we do in WT due to jobs doing that is a bit unfortunate and we try to avoid it.
However, I will make it a vector with 512 reserved slots. It will cost nothing up to that point and then a chunk will be allocated by stl.

jobmanager/WorkerThread.cpp
```diff
@@ -2,39 +2,89 @@
...
+
         while (_running)
         {
-            auto job = _jobManager.wait();
-            if (!job)
+            auto jobProcessed = processJobs();
+            if (!jobProcessed || _backgroundJobCount == _backgroundJobs.size())
             {
-                continue;
+                pollInterval = std::min(maxWait2ms, pollInterval * 2);
```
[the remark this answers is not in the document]
olofkallander: you are right

jobmanager/WorkerThread.cpp
```diff
@@ -2,14 +2,35 @@
...
 {
 
 WorkerThread::WorkerThread(jobmanager::JobManager& jobManager)
     : _running(true),
       _jobManager(jobManager),
+      _backgroundJobCount(0),
       _thread([this] { this->run(); })
 {
+    _backgroundJobs.reserve(512);
+    std::fill(_backgroundJobs.begin(), _backgroundJobs.end(), nullptr);
```
RicardoMDomingues, line 22: This is doing nothing. Reserve only reserve space internally on memory managed by std::vector. But it change the size (the elements that actually being used by vector). Then  in this point `backgroundJobs.begin()` is equal to `_backgroundJobs.end()`

RicardoMDomingues (approved):
LTGM. just minor comment

olofkallander:
> This is doing nothing. Reserve only reserve space on internally memory managed by std::vector. But it does not change the size (the elements that actually stored in vector). Then  in this point `backgroundJobs.begin()` is equal to `_backgroundJobs.end()` 

true, was too much on my mind.

test/bridge/ActiveMediaListTest.cpp
```diff
@@ -18,17 +18,6 @@ namespace
 static const uint32_t defaultLastN = 2;
 static const uint32_t audioLastN = 3;
```
olofkallander, line 20: This thread could grab all pending jobs and exit before we actually posted the StopJob on JobQueue. Then the ~JobQueue would stall forever. That happened in Jenkins all of a sudden. This test has been like this for over a year.

PR #168, opened 2022-09-06
races

olofkallander:


test/integration/IntegrationTest.cpp
```diff
@@ -39,35 +39,35 @@
...
 
+// TimeTurner time source must be set before starting any threads.
+// Fake internet thread, JobManager timer thread, worker threads.
 void IntegrationTest::SetUp()
 {
-
     if (__has_feature(address_sanitizer) || __has_feature(thread_sanitizer))
     {
-        GTEST_SKIP();
+        // GTEST_SKIP();
```
olofkallander, line 57: This will be disabled again until we find a reliable way of running in Jenkins.

bridge/engine/VideoForwarderRewriteAndSendJob.cpp
```diff
@@ -65,26 +64,27 @@ void VideoForwarderRewriteAndSendJob::run()
...
     {
+        assert(false); // should have been checked multiple times
         return;
     }
 
-    bool isRetransmittedPacket = false;
     if (_outboundContext.rtpMap.format == RtpMap::Format::VP8RTX)
     {
-        isRetransmittedPacket = true;
+        logger::warn("%s rtx packet should not reach rewrite and send. ssrc %u, seq %u",
```
olofkallander, line 73: See https://perzoinc.atlassian.net/browse/RTC-13340
rtx should be converted to (out of order rtp packets). Payload type should never be rtx once processed in RtxReceiveJob

bridge/engine/VideoForwarderRtxReceiveJob.cpp
```diff
@@ -64,22 +67,19 @@ void VideoForwarderRtxReceiveJob::run()
...
-    _ssrcFeedbackContext.lastUnprotectedExtendedSequenceNumber = _extendedSequenceNumber;
-    Vp8Rewriter::rewriteRtxPacket(*_packet, _mainSsrc);
-
-    if (!_ssrcFeedbackContext.videoMissingPacketsTracker.get())
-    {
-        assert(false);
-        return;
-    }
+    _rtxSsrcContext.lastUnprotectedExtendedSequenceNumber = _extendedSequenceNumber;
+    Vp8Rewriter::rewriteRtxPacket(*_packet,
```
olofkallander, line 76: rewriter will turn this into vp8 rtp packet including payloadtype.
Please speak up if you think this is wrong.

test/integration/IntegrationTest.cpp
reddvl1980: Shall we change it back/disable before merging?

math/Fields.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include <cstddef>
+#include <type_traits>
+
+namespace math
+{
+
+// difference (b - a) mod (1<<BITS)
+template <typename T, size_t BITS>
+constexpr typename std::make_signed<T>::type ringDifference(T a, const T b)
```
reddvl1980, line 10: Good that you've added unit tests for this!

test/bridge/Vp8RewriterTest.cpp
```diff
@@ -29,16 +58,31 @@ class Vp8RewriterTest : public ::testing::Test
...
+
+    {
+        const auto offset = math::ringDifference<uint32_t, 12u>(0xFFF, 444);
+        EXPECT_EQ(offset, 445);
+    }
+
+    {
+        const auto offset = math::ringDifference<uint16_t, 16u>(0xFFFF, 444);
+        EXPECT_EQ(offset, 445);
+    }
```
reddvl1980, line 80: Consider reinforcing tests for conventional "rings" (e.g. 16 bits) by checking that, say,  offset + 444 == 0xffff.

bridge/engine/VideoForwarderRtxReceiveJob.cpp
reddvl1980: @RicardoMDomingues, as we spoke with Olof, it this will disable RTX forwarding. Instead, all RTX packets would be re-transmitted (forwarded) as non-RTX, "normal" VP8 RTP packets on a "main" ssrc (not on a feedback). That, in theory, should get rid of domino effect when clients needs some packets, which was lost on a barbell link. Instead of forwarded RTX, client will receive normal, but out-of-order RTP. That should be ok for the client, even that formally it's NACK requewst was not served.

bridge/engine/EngineStreamDirector.h
```diff
@@ -69,12 +70,12 @@ class EngineStreamDirector
         if (_participantStreams.find(endpointIdHash) != _participantStreams.end())
         {
             logger::debug("addParticipant stream already added, endpointIdHash %lu",
```
reddvl1980, line 72: @olofkallander, Do we want to keep these explicit logger or change them to DIRECTOR_LOG?

math/Fields.h
```diff
@@ -0,0 +1,26 @@
...
+{
+    static_assert(std::is_unsigned<T>::value, "Difference only works on unsigned integers");
+    using SignedT = typename std::make_signed<T>::type;
+    if (a & (1 << (BITS - 1)))
```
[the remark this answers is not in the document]
reddvl1980: @RicardoMDomingues , as far as I understood this should work _only_ with smaller "rings". E.g. 63 bits and smalller. So in your example 'a' could be uint32_t.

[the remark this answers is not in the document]
RicardoMDomingues: Not sure if understood. `a` can be a `uint64_t` and BITS to be 63? That will also cause issues. The expression `(1 << (BITS - 1))` will cause a compiler warning each time `BITS >= 32` as shift count is bigger than number of int bits. So in the end it will be truncated to int before be converted to uint64_t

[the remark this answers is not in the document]
olofkallander: don't worry I am working on it. I also see cases were it is not producing what I expect.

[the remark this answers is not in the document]
reddvl1980: What I meant is that value of `a` should be expressed by BITS number of bits. So in your example value of `a` should be 0x000000007FFFFFFF.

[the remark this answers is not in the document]
RicardoMDomingues: Even if  this work for our current cases this accept a generic types. One day someone could find this function useful and use with a combination hat does not work correctly and it will create problem that are very hard to find

bridge/engine/VideoForwarderRtxReceiveJob.cpp
RicardoMDomingues: 👍

PR #169, opened 2022-09-07
RTC-12935 Fix video RTX

reddvl1980:
- Create an integration test that uses network simulation to control packet losses over the barbell connection;
- Add tracking of the missing video packets to the test client (SfuClient);
- Make EningeMixer to process missing video packets that came over barbell;
- Process NACK requests that comes via barbell;
- Fix the bug in the VideoForwarderRtxReceiveJob that kept received video packet on feedback ssrc context.

### DIFF
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -2630,23 +2630,27 @@ void EngineMixer::processIncomingTransportFbRtcpPacket(const transport::RtcTrans
 
     if (fromBarbell)
     {
-        const auto bb = _engineBarbells.getItem(transport->getEndpointIdHash());
-        if (bb)
+        const auto barbell = _engineBarbells.getItem(transport->getEndpointIdHash());
+        if (barbell)
         {
             uint32_t feedbackSsrc = 0;
             _activeMediaList->getFeedbackSsrc(mediaSsrc, feedbackSsrc);
 
-            auto& bbTransport = bb->transport;
-            auto* mediaSsrcOutboundContext =
-                obtainOutboundSsrcContext(bb->idHash, bb->ssrcOutboundContexts, mediaSsrc, bb->audioRtpMap);
+            auto& bbTransport = barbell->transport;
+            auto* mediaSsrcOutboundContext = obtainOutboundSsrcContext(barbell->idHash,
+                barbell->ssrcOutboundContexts,
+                mediaSsrc,
+                barbell->audioRtpMap);
             if (!mediaSsrcOutboundContext || !mediaSsrcOutboundContext->packetCache.isSet() ||
                 !mediaSsrcOutboundContext->packetCache.get())
             {
                 return;
             }
 
-            auto* feedbackSsrcOutboundContext =
-                obtainOutboundSsrcContext(bb->idHash, bb->ssrcOutboundContexts, feedbackSsrc, bb->audioRtpMap);
+            auto* feedbackSsrcOutboundContext = obtainOutboundSsrcContext(barbell->idHash,
+                barbell->ssrcOutboundContexts,
+                feedbackSsrc,
+                barbell->audioRtpMap);
             if (!feedbackSsrcOutboundContext)
             {
                 return;
@@ -3572,15 +3576,15 @@ void EngineMixer::processEngineMissingPackets(bridge::SsrcInboundContext& ssrcIn
 
 void EngineMixer::processBarbellMissingPackets(bridge::SsrcInboundContext& ssrcInboundContext)
 {
-    const auto bb = _engineBarbells.getItem(ssrcInboundContext.sender->getEndpointIdHash());
-    if (bb)
+    const auto barbell = _engineBarbells.getItem(ssrcInboundContext.sender->getEndpointIdHash());
+    if (barbell)
     {
-        auto videoStream = bb->videoSsrcMap.getItem(ssrcInboundContext.ssrc);
+        auto videoStream = barbell->videoSsrcMap.getItem(ssrcInboundContext.ssrc);
         if (videoStream)
         {
-            bb->transport.getJobQueue().addJob<bridge::ProcessMissingVideoPacketsJob>(ssrcInboundContext,
+            barbell->transport.getJobQueue().addJob<bridge::ProcessMissingVideoPacketsJob>(ssrcInboundContext,
                 0,
-                bb->transport,
+                barbell->transport,
                 _sendAllocator);
             return;
         }
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -1389,7 +1389,8 @@ void EngineMixer::onVideoRtpRtxPacketReceived(SsrcInboundContext* ssrcContext,
     const auto isSenderInLastNList = _activeMediaList->isInActiveVideoList(endpointIdHash);
     const bool mustBeForwardedOnBarbells = isSenderInLastNList && !_engineBarbells.empty() && !isFromBarbell;
 
-    if (!mustBeForwardedOnBarbells &&
+    // Optmization with isSsrcUsed is not need for barbell (barbell sends only necessary streams).
+    if (videoStream && !mustBeForwardedOnBarbells &&
         !_engineStreamDirector->isSsrcUsed(mainSsrc,
             videoStream->endpointIdHash,
             isSenderInLastNList,
@@ -1397,6 +1398,11 @@ void EngineMixer::onVideoRtpRtxPacketReceived(SsrcInboundContext* ssrcContext,
     {
         return;
     }
+
+    logger::debug("!!! onVideoRtpRtxPacketReceived : %c sender %s",
+        _loggableId.c_str(),
+        isFromBarbell ? 't' : 'f',
+        sender->getLoggableId().c_str());
 
     sender->getJobQueue().addJob<bridge::VideoForwarderRtxReceiveJob>(std::move(packet),
         sender,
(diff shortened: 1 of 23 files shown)


test/integration/BarbellTest.cpp
```diff
@@ -0,0 +1,282 @@
...
+            {
+                if (group.clients[0]->isRemoteVideoSsrc(item.first))
+                {
+                    continue;
+                }
+
+                std::vector<double> freqVector;
+                std::vector<std::pair<uint64_t, double>> amplitudeProfile;
+                auto rec = item.second->getRecording();
+                analyzeRecording(rec, freqVector, amplitudeProfile, item.second->getLoggableId().c_str());
```
reddvl1980, line 247: @olofkallander, @RicardoMDomingues, do you think we need this audio analisys in this kind of test? It takes some time (~5 s) to analyze, making overall test execution slower. But in the same time, it guarantees that the data we receive is intact.

RicardoMDomingues: imo 5sec is not a big deal. But I guess testing if the audio is intact is only necessary for participants with mixed audio where smb needs to decode, mix and encode again, otherwise as sfu we are interested only in understanding if we forwarder the packets as we should and we should be able do that without analyse the audio it self

RicardoMDomingues: imo 5sec is not a big deal. But I guess testing if the audio is intact is only necessary for participants with mixed audio where smb needs to decode, mix and encode again, otherwise as sfu we are interested only in understanding if we forwarder the packets as we should and we should be able do that without analyse the audio it self

reddvl1980: OK. I'll keep it here. But I'll skip it in the next tests where we'll check packet losses and video recovery.

test/integration/BarbellTest.cpp
```diff
@@ -0,0 +1,282 @@
...
+        "ice.singlePort":12000,
+        "port":8090,
+        "recording.singlePort":12500,
+        "rctl.enable": false
+        })");
+
+        auto bridge2 = std::make_unique<bridge::Bridge>(config2);
+        bridge2->initialize(_bridgeEndpointFactory);
+
+        ScopedFinalize finalize(std::bind(&IntegrationTest::finalizeSimulation, this));
```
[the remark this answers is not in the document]
reddvl1980: Good catch!

bridge/engine/EngineMixer.cpp
```diff
@@ -2624,8 +2625,54 @@ void EngineMixer::processIncomingTransportFbRtcpPacket(const transport::RtcTrans
         return;
     }
 
+    const auto fromBarbell = EngineBarbell::isFromBarbell(transport->getTag());
     const auto mediaSsrc = rtcpFeedback->mediaSsrc.get();
 
+    if (fromBarbell)
+    {
+        const auto bb = _engineBarbells.getItem(transport->getEndpointIdHash());
```
olofkallander, line 2633: Perhaps we should just spell it out barbell instead of doing abbreviations.

test/integration/BarbellTest.cpp
```diff
@@ -0,0 +1,296 @@
...
+            logTransportSummary("client3", group.clients[2]->_transport.get(), transportSummary3);
+
+            EXPECT_NEAR(videoReceiveStats.packets,
+                transportSummary2.begin()->second.packetsSent + transportSummary3.begin()->second.packetsSent,
+                25);
+        }
+    });
+}
+
+#undef DEFINE_3_CLIENT_CONFERENCE
```
olofkallander, line 296: can remove

test/integration/IntegrationTest.h
```diff
@@ -13,6 +14,27 @@
 #include <gtest/gtest.h>
 #include <mutex>
 
+#define DEFINE_3_CLIENT_CONFERENCE(TChannel, BASE_URL)                                                                 \
```
olofkallander, line 17: we do not use this

test/integration/BarbellTest.cpp
```diff
@@ -0,0 +1,296 @@
+#include "BarbellTest.h"
+
+#include "api/ConferenceEndpoint.h"
```
olofkallander, line 3: We should check which of these includes are needed. It looks like a few may not be needed.

test/integration/emulator/SfuClient.h
```diff
@@ -37,6 +40,41 @@ class MediaSendJob : public jobmanager::Job
...
+    size_t sndPacketsSent = 0;
+
+    RtxStats& operator+=(const RtxStats& other)
+    {
+        rcvPacketsMissing += other.rcvPacketsMissing;
+        rcvPacketsRecovered += other.rcvPacketsRecovered;
+        rcvNackRequestSent += other.rcvNackRequestSent;
+
+        sndNackRequestsReceived += other.sndNackRequestsReceived;
+        sndPacketsMissingAsked += other.sndPacketsMissingAsked;
```
olofkallander, line 63: what does this mean?

test/integration/emulator/SfuClient.h
```diff
@@ -37,6 +40,41 @@ class MediaSendJob : public jobmanager::Job
...
+        sndPacketsMissingSent += other.sndPacketsMissingSent;
```
olofkallander, line 64: what

test/integration/emulator/SfuClient.h
```diff
@@ -37,6 +40,41 @@ class MediaSendJob : public jobmanager::Job
     memory::UniquePacket _packet;
 };
 
+struct RtxStats
+{
+    // Receiver's stats: how much we missed, recovered and complained.
+    size_t rcvPacketsMissing = 0;
+    size_t rcvPacketsRecovered = 0;
+    size_t rcvNackRequestSent = 0; //  Not implemented: SfuClient does not sent NACK yet.
```
olofkallander, line 48: received sent ?

bridge/engine/EngineMixer.cpp
```diff
@@ -2624,8 +2625,54 @@ void EngineMixer::processIncomingTransportFbRtcpPacket(const transport::RtcTrans
...
+                    bbTransport,
+                    *(mediaSsrcOutboundContext->packetCache.get()),
+                    pid,
+                    blp,
+                    feedbackSsrc,
+                    timestamp,
+                    transport->getRtt());
+            }
+        }
+        return;
```
RicardoMDomingues, line 2673: This is quite extensive method and in the end has an early return. I would prefer having the early returns not far way from the condition. Could you put this in other method. like:
```
if (fromBarbell)
{
    processIncomingFbRtcpPacketFromBarbellTransport(transport, rtcpPacket, timestamp);
    return;
}

File: bridge/engine/EngineMixer.cpp [hunk 11]
```diff
@@ -3510,6 +3557,36 @@ void EngineMixer::allocateRecordingRtpPacketCacheIfNecessary(SsrcOutboundContext
...
+    if (bb)
+    {
+        auto videoStream = bb->videoSsrcMap.getItem(ssrcInboundContext.ssrc);
+        if (videoStream)
+        {
+            bb->transport.getJobQueue().addJob<bridge::ProcessMissingVideoPacketsJob>(ssrcInboundContext,
+                0,
+                bb->transport,
+                _sendAllocator);
+            return;
```
RicardoMDomingues, line 3585: Unnecessary

test/integration/emulator/SfuClient.h
olofkallander: when we have a group of metrics like this it is better to put them in named or unnamed struct in a variable called receive or rcv so we reference them by rcv.packetsMissing

bridge/engine/EngineMixer.cpp
```diff
@@ -3510,6 +3557,36 @@ void EngineMixer::allocateRecordingRtpPacketCacheIfNecessary(SsrcOutboundContext
...
+void EngineMixer::processBarbellMissingPackets(bridge::SsrcInboundContext& ssrcInboundContext)
+{
+    const auto bb = _engineBarbells.getItem(ssrcInboundContext.sender->getEndpointIdHash());
+    if (bb)
+    {
+        auto videoStream = bb->videoSsrcMap.getItem(ssrcInboundContext.ssrc);
+        if (videoStream)
+        {
+            bb->transport.getJobQueue().addJob<bridge::ProcessMissingVideoPacketsJob>(ssrcInboundContext,
+                0,
```
RicardoMDomingues, line 3582: Magic number. Put at least a comment to know the parameter name

test/integration/emulator/SfuClient.h
```diff
@@ -375,6 +448,86 @@ class SfuClient : public transport::DataReceiver
...
+
+                inboundContext.lastReceivedExtendedSequenceNumber = extendedSequenceNumber;
+            }
+            else if (extendedSequenceNumber != inboundContext.lastReceivedExtendedSequenceNumber)
+            {
+                uint32_t esn = 0;
+                if (!inboundContext.videoMissingPacketsTracker->onPacketArrived(sequenceNumber, esn) ||
+                    esn != extendedSequenceNumber)
+                {
+                    logger::debug("%s Unexpected retranmsission packet seq %u ssrc %u, dropping",
```
olofkallander, line 496: re-transmission

test/integration/BarbellTest.cpp
```diff
@@ -0,0 +1,296 @@
...
+            3);
+
+        Conference conf;
+        group.startConference(conf, baseUrl);
+
+        Conference conf2;
+        group.startConference(conf2, baseUrl2);
+
+        ScopedFinalize finalize(std::bind(&IntegrationTest::finalizeSimulation, this));
+        startSimulation();
```
RicardoMDomingues, line 173: Shouldn't it be placed after `Barbell` object creation?

test/integration/IntegrationTest.h
```diff
@@ -35,7 +57,8 @@ struct IntegrationTest : public ::testing::Test
 
     std::unique_ptr<transport::TransportFactory> _transportFactory;
     std::shared_ptr<fakenet::InternetRunner> _internet;
-    std::shared_ptr<transport::EndpointFactory> _endpointFacory;
+    std::shared_ptr<transport::EndpointFactory> _bridgeEndpointFactory;
+    std::shared_ptr<transport::EndpointFactory> _clientsEndpointFacory;
```
RicardoMDomingues, line 61: Typo. Factory

olofkallander (approved):

PR #172, opened 2022-09-13
Automatically remove idle streams

reddvl1980:
C9 feature request: automatically remove idle endpoints after configurable timeout.
Idle endpoints are those which did not receive no packets (ICE, DTLS, DTP, RTCP).


RicardoMDomingues (commented):
This should be archived by endpoint configuration otherwise we will break SIP recording as we they are send only transports (Not sure if we receive RTCP but I am guessing we did not).

Also for the Symphony perspective this could cause a disagreement of MBR and CS. It would be more logic remove by ICE failing rather than media timeout. If there is no media but we still have ICE, probably the clients is alive but removing should be somehow a decision that CS needs to take. Then this only seems to fit a C9 specific requirement that I think is not the best for Symphony and probably will not work of SIP records, then my suggestion make it configure on endpoint allocation request

reddvl1980:
@RicardoMDomingues , @olofkallander , as agreed, endpoint idle timeout is now set via allocateEndpoint API and stays constant. Config file changes are removed.

olofkallander (approved):

api/Parser.cpp
```diff
@@ -294,6 +294,11 @@ AllocateEndpoint parseAllocateEndpoint(const nlohmann::json& data)
         allocateEndpoint._data.set(AllocateEndpoint::Data());
     }
 
+    if (data.find("idleTimeout") != data.end())
+    {
+        setIfExists(allocateEndpoint._idleTimeoutSeconds, data, "idleTimeout");
+    }
```
RicardoMDomingues, line 300: `setExists` has name suggest already verify  this: `data.find("idleTimeout") != data.end())`
```suggestion
 setIfExists(allocateEndpoint._idleTimeoutSeconds, data, "idleTimeout");
```

bridge/engine/EngineMixer.cpp
```diff
@@ -3834,15 +3838,57 @@ void EngineMixer::checkIfRateControlIsNeeded(const uint64_t timestamp)
...
+                "EngineMixer",
+                stream->endpointIdHash,
+                lastReceivedTimestamp,
+                timestamp);
+            removeMethod(stream);
+        }
+    }
+}
+
+#define REMOVE_IDLE_STREAMS(type)                                                                                      \
```
RicardoMDomingues, line 3872: Macro is something we should avoid as mush as we can on modern C++. I am sure we are able to similar without using templates

bridge/engine/EngineMixer.cpp
```diff
@@ -3834,15 +3838,57 @@ void EngineMixer::checkIfRateControlIsNeeded(const uint64_t timestamp)
...
+    ::bridge::removeIdleStreams<Engine##type##Stream>(_engine##type##Streams,                                          \
+        std::bind(&EngineMixer::remove##type##Stream, this, std::placeholders::_1),                                    \
```
RicardoMDomingues, line 3874: `std::bind` most probably will create a dynamic memory which is something we don't allow on RT thread. There is a recommendation to std::function to have a min storage to have at least space to a pointer without a dynamic allocation but even that is not ensured by standard although we now most used compilers will not allocate if you capture 8 bytes on a lambda expression.

But as far as I know `std::bind` due to the complexity of implementation always use dynamic memory.

RicardoMDomingues (changes requested):

bridge/engine/EngineMixer.cpp
reddvl1980: Removed.

bridge/engine/EngineMixer.cpp
reddvl1980: Removed lambda/binding and instead overloaded removeStream(...) in mixer.

bridge/engine/EngineMixer.h
```diff
@@ -97,14 +97,14 @@ class EngineMixer : public transport::DataReceiver
 
     // -- methods executed on engine thread
     void addAudioStream(EngineAudioStream* engineAudioStream);
-    void removeAudioStream(EngineAudioStream* engineAudioStream);
+    void removeStream(EngineAudioStream* engineAudioStream);
```
olofkallander, line 100: good

bridge/engine/EngineMixer.cpp
```diff
@@ -3834,13 +3845,48 @@ void EngineMixer::checkIfRateControlIsNeeded(const uint64_t timestamp)
...
+{
+    if (utils::Time::diffLT(_lastIdleTransportCheck, timestamp, 1ULL * utils::Time::sec))
+    {
+        return;
     }
+    _lastIdleTransportCheck = timestamp;
+
+    ::bridge::removeIdleStreams<EngineVideoStream>(_engineVideoStreams, this, timestamp);
+    ::bridge::removeIdleStreams<EngineAudioStream>(_engineAudioStreams, this, timestamp);
+    ::bridge::removeIdleStreams<EngineDataStream>(_engineDataStreams, this, timestamp);
```
RicardoMDomingues, line 3889: [minor] we are on bridge namespace context already. Why do we need to specify all path.

Also TStream can be auto deducted by compiler. Only this should do enough:
`removeIdleStreams(_engineDataStreams, this, timestamp);`

bridge/engine/EngineMixer.cpp
```diff
@@ -3834,13 +3845,48 @@ void EngineMixer::checkIfRateControlIsNeeded(const uint64_t timestamp)
...
+
+void EngineMixer::removeIdleStreams(const uint64_t timestamp)
+{
+    if (utils::Time::diffLT(_lastIdleTransportCheck, timestamp, 1ULL * utils::Time::sec))
+    {
+        return;
     }
+    _lastIdleTransportCheck = timestamp;
+
+    ::bridge::removeIdleStreams<EngineVideoStream>(_engineVideoStreams, this, timestamp);
```
olofkallander, line 3887: It looks like if you put template param on container type instead, you do not need to specify the type at all here. It will automagically resolve in the removeIdleStreams method.

bridge/engine/EngineMixer.cpp
```diff
@@ -3834,13 +3845,48 @@ void EngineMixer::checkIfRateControlIsNeeded(const uint64_t timestamp)
     _probingVideoStreams = enableBEProbing;
 }
 
-void EngineMixer::runTransportTicks(const uint64_t timestamp)
+template <typename TStream>
+void removeIdleStreams(concurrency::MpmcHashmap32<size_t, TStream*>& streams,
```
RicardoMDomingues, line 3849: [minor] I noticed now this is a free function. If it's a free function it should go to anonymous namespace. But as we receive the pointer to `EngineMixer* mixer` perhaps it make more sense to be member of EngineMixer instead. There is no problem in having a private templated member declared in .h file and the definition on cpp as the function will be only used by the compile unit that has the definition

RicardoMDomingues (commented):
LGTM. Just minor comments

RicardoMDomingues (approved):

bridge/engine/EngineMixer.cpp
```diff
@@ -583,6 +584,13 @@ void EngineMixer::removeDataStream(EngineDataStream* engineDataStream)
         endpointIdHash);
 
     _engineDataStreams.erase(endpointIdHash);
+
+    EngineMessage::Message message(EngineMessage::Type::DataStreamRemoved);
+    message.command.dataStreamRemoved.mixer = this;
+    message.command.dataStreamRemoved.engineStream = engineDataStream;
+    engineDataStream->transport.getJobQueue().addJob<SendEngineMessageJob>(engineDataStream->transport,
+        _messageListener,
+        std::move(message));
```
RicardoMDomingues, line 593: Good catch 👍

PR #175, opened 2022-09-15
Config HTTP bind address and use localhost by default

RicardoMDomingues:


reddvl1980 (approved):

olofkallander (approved):
We should enhance parser to take port as well. For another day.

PR #179, opened 2022-09-16
disable int tests asan , lcov builds

olofkallander:


test/integration/IntegrationTest.cpp
```diff
@@ -35,6 +35,12 @@
 
 #define USE_FAKENETWORK 1
 
+#if (DEBUG || !(__has_feature(address_sanitizer) || __has_feature(thread_sanitizer) || LCOV_BUILD))
```
olofkallander, line 38: It does not work yet. Will fix

RicardoMDomingues: I suggest to define `ENABLE_INT_TESTS` on cmake level. So our code can be a bit less clang dependent as (__has_feature is a clang function)

PR #181, opened 2022-09-19
Update examples to new API

marcusspangenberg:
- Update examples/simpleserver to use new API
- Update dependencies of simpleserver and simpleclient

### DIFF
--- a/examples/simpleserver/src/main/java/com/symphony/simpleserver/smb/api/Parser.java
+++ b/examples/simpleserver/src/main/java/com/symphony/simpleserver/smb/api/Parser.java
@@ -7,6 +7,7 @@ import com.symphony.simpleserver.sdp.objects.*;
 import org.springframework.stereotype.Component;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 
 @Component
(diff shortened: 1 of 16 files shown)


examples/simpleserver/src/main/java/com/symphony/simpleserver/smb/api/Parser.java
```diff
@@ -377,17 +380,38 @@ public SmbEndpointDescription makeEndpointDescription(SessionDescription sdpAnsw
...
-                    video.ssrcGroups.add(smbSsrcGroup);
+                final var streamsMap = new HashMap<String, SmbVideoStream>();
+
+                mediaDescription.ssrcs.forEach(ssrc -> {
+                    final var smbVideoStream = streamsMap.computeIfAbsent(
+                            ssrc.mslabel,
+                            key -> {
+                                final var value = new SmbVideoStream();
+                                value.id = ssrc.mslabel;
+                                value.content = "video";
```
RicardoMDomingues, line 391: [Minor] Probably we don't support screensharing on simpleclient, but I guess we can handle it on simple server as example how it should be handled

```suggestion
                                value.content = "slides".equals(mediaDescription.content)?  "slides" : "video";
```

RicardoMDomingues (approved):
LGTM. Thanks for update the examples 👍

PR #186, opened 2022-09-20
Make FFT analysis multi-threaded

reddvl1980:


test/integration/IntegrationTest.cpp
```diff
@@ -223,6 +224,16 @@ api::ConferenceEndpointExtendedInfo IntegrationTest::getEndpointExtendedInfo(con
     return api::Parser::parseEndpointExtendedInfo(endpointRequest.getJsonBody());
 }
 
+size_t getNumWorkingThreads()
```
olofkallander, line 227: This name in the same file creates confusion as there is already a getNumWorkerThreads.

test/integration/IntegrationTest.cpp
```diff
@@ -256,26 +266,51 @@ void IntegrationTest::analyzeRecording(const std::vector<int16_t>& recording,
...
+    std::mutex fft_mutex;
 
-        std::vector<double> frequencies;
-        SampleDataUtils::listFrequencies(testVector, codec::Opus::sampleRate, frequencies);
-        for (size_t i = 0; i < frequencies.size() && i < 50; ++i)
-        {
-            if (std::find(frequencyPeaks.begin(), frequencyPeaks.end(), frequencies[i]) == frequencyPeaks.end())
+    for (size_t threadId = 0; threadId < numThreads; threadId++)
+    {
+        workers.push_back(std::thread([&frequencyPeaks, &recording, &fft_mutex, &logId, size, numThreads, threadId]() {
```
olofkallander, line 277: I have read multiple times. I still do not like that we create a fat lambda starting at 3 indentation levels.
Break this out into a method and pass that one to the thread.

test/integration/IntegrationTest.cpp
reddvl1980: Agree.

test/integration/IntegrationTest.cpp
```diff
@@ -256,26 +266,51 @@ void IntegrationTest::analyzeRecording(const std::vector<int16_t>& recording,
...
+                    testVector[x] =
+                        std::complex<double>(static_cast<double>(recording[x + cursor]), 0.0) / (256.0 * 128);
+                }
+
+                SampleDataUtils::fft(testVector);
+
+                std::vector<double> frequencies;
+                SampleDataUtils::listFrequencies(testVector, codec::Opus::sampleRate, frequencies);
+                {
+                    const std::lock_guard<std::mutex> lock(fft_mutex);
```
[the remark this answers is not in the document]
reddvl1980: Fixed (with single threaded reduction in the end, do not think making reduction multithreaded is necessary).

Issue #187, opened 2022-09-21
Transport::runTicks is no longer called


marcusspangenberg:
In PR #172 EngineMixer::runTransportTicks was removed. This means Transport::runTick() is never called anywhere anymore. 

Was this the intent of the change, otherwise it seems like a possible issue. If runTick() is never to be used anymore, the function in Transport should be removed. Right now it just looks wrong.

marcusspangenberg:
Ok, seems like EngineMixer::runTransportTicks was never called anywhere either. So should Transport::runTick be called ever or removed?

reddvl1980:
I think it needs to be removed. @olofkallander, remember I showed it to you, what do you think?

reddvl1980:
https://github.com/finos/SymphonyMediaBridge/pull/190

PR #190, opened 2022-09-22
Run transport maintenance (e.g. drain pacing buffer) periodically

reddvl1980:
Issue: https://github.com/finos/SymphonyMediaBridge/issues/187

### DIFF
--- a/transport/TransportImpl.cpp
+++ b/transport/TransportImpl.cpp
@@ -379,12 +379,22 @@ private:
 class RunTickJob : public jobmanager::CountedJob
 {
 public:
-    RunTickJob(TransportImpl& transport) : CountedJob(transport.getJobCounter()), _transport(transport) {}
+    RunTickJob(TransportImpl& transport, const uint64_t timestamp)
+        : CountedJob(transport.getJobCounter()),
+          _transport(transport),
+          _timestamp(timestamp)
+    {
+    }
 
-    void run() override { _transport.doRunTick(utils::Time::getAbsoluteTime()); }
+    void run() override
+    {
+        _transport._lastTickJobStartTimestamp = _timestamp;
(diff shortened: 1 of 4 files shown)


transport/TransportImpl.cpp
```diff
@@ -379,12 +379,22 @@ class UnregisteredJob : public jobmanager::CountedJob
...
+        : CountedJob(transport.getJobCounter()),
+          _transport(transport),
+          _timestamp(timestamp)
+    {
+    }
 
-    void run() override { _transport.doRunTick(utils::Time::getAbsoluteTime()); }
+    void run() override
+    {
+        _transport._lastTickJobStartTimestamp = _timestamp;
```
RicardoMDomingues, line 391: `_lastTickJobStartTimestamp` should be updated before post the new job on JobManager, otherwise it can be posted multiple times if smb starving on Worker threads

transport/TransportImpl.cpp
```diff
@@ -2448,9 +2459,13 @@ void TransportImpl::setRtxProbeSource(const uint32_t ssrc, uint32_t* sequenceCou
 
 void TransportImpl::runTick(uint64_t timestamp)
 {
+    if (utils::Time::diffLT(_lastTickJobStartTimestamp, timestamp, utils::Time::ms * 100))
```
olofkallander, line 2462: I would move the check to enginemixer to avoid looping through streams every 10ms.

transport/TransportImpl.cpp
olofkallander: removing _lastTickJobStartTimestamp and doing the check in EngineMixer resolves that and reduces iterating through all video transports.

olofkallander (approved):

olofkallander:
you hit the tcheck race that my packet cache race fixes

PR #192, opened 2022-09-22
Olofk/pcacherace

olofkallander:
SsrcOutboundContext and packetcache caused multiple races.
Fixed by moving some operations onto Transport Jobs and some shared items made atomic.


bridge/engine/RecordingVideoForwarderSendJob.cpp
```diff
@@ -0,0 +1,147 @@
...
+void RecordingVideoForwarderSendJob::run()
+{
+    auto rtpHeader = rtp::RtpHeader::fromPacket(*_packet);
+    if (!rtpHeader)
+    {
+        assert(false); // should have been checked multiple times
+        return;
+    }
+
+    if (_outboundContext.rtpMap.format == RtpMap::Format::VP8RTX)
```
olofkallander, line 43: Is this actually possible for recording streams?

bridge/engine/RecordingVideoForwarderSendJob.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+            logger::debug("%s requested key frame from %u on ssrc %u",
+                "RecordingVideoForwarderSendJob",
+                _transport.getLoggableId().c_str(),
+                _senderInboundContext.ssrc,
+                _outboundContext.ssrc);
+        }
+    }
+
+    if (!_outboundContext.shouldSend(rtpHeader->ssrc, _extendedSequenceNumber))
```
olofkallander, line 102: This is where we check if the packet happens to be too old (512 pkts). It also always allows a new source ssrc to pass through as it means new source stream for rewrite. I guess for recording that rewrite is not actually happening.

bridge/engine/SsrcOutboundContext.h
```diff
@@ -99,32 +95,22 @@ class SsrcOutboundContext
...
+    {
+        lastSendTime = timestamp;
+        idle = false;
+    }
 
     // Stream owner is being removed. Stop outbound packets over this context
     bool markedForDeletion;
-
     // Stream is temporarily idle and srtp encryption context is removed. It may be activated again.
     bool idle;
```
olofkallander, line 113: I removed the PLI metric that appears only in logs. It was a source for data race.

bridge/engine/RecordingAudioForwarderSendJob.cpp
```diff
@@ -53,11 +60,25 @@ void RecordingAudioForwarderSendJob::run()
         {
             if (!packetCache->add(*_packet, nextSequenceNumber))
             {
-                logger::info("Dropping rec audio packet - cache issues...", "RecordingAudioForwarderSendJob");
-                return;
+                logger::warn("Failed to cache rec audio packet", "RecordingAudioForwarderSendJob");
```
olofkallander, line 63: let the audio packet through also when packet cache is not accepting it.

RicardoMDomingues: Yes, that make sense 👍

RicardoMDomingues (approved):

PR #195, opened 2022-09-23
Olofk/cleanup

olofkallander:
around missing packets tracker.

### DIFF
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -988,7 +988,7 @@ void EngineMixer::processMissingPackets(const uint64_t timestamp)
         }
     }
 
-    processRecordingMissingPackets(timestamp);
+    processRecordingUnackedPackets(timestamp);
 }
 
 void EngineMixer::runDominantSpeakerCheck(const uint64_t engineIterationStartTimestamp)
@@ -3646,7 +3646,7 @@ void EngineMixer::processBarbellMissingPackets(bridge::SsrcInboundContext& ssrcI
     }
 }
 
-void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
+void EngineMixer::processRecordingUnackedPackets(const uint64_t timestamp)
 {
     for (auto& engineRecordingStreamEntry : _engineRecordingStreams)
     {
@@ -3654,10 +3654,6 @@ void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
         for (auto& recEventMissingPacketsTrackerEntry : engineRecordingStream->recEventUnackedPacketsTracker)
         {
             auto& recEventMissingPacketsTracker = recEventMissingPacketsTrackerEntry.second;
-            if (!recEventMissingPacketsTracker.shouldProcess(timestamp / 1000000ULL))
-            {
-                continue;
-            }
 
             auto transportItr = engineRecordingStream->transports.find(recEventMissingPacketsTrackerEntry.first);
             if (transportItr == engineRecordingStream->transports.end())
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -52,7 +52,7 @@ namespace
 {
 
 const int16_t mixSampleScaleFactor = 4;
-const uint64_t RECUNACK_PROCESS_INTERVAL = 50 * utils::Time::ms;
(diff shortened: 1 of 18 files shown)


bridge/engine/EngineMixer.cpp
```diff
@@ -3646,18 +3646,14 @@ void EngineMixer::processBarbellMissingPackets(bridge::SsrcInboundContext& ssrcI
     }
 }
 
-void EngineMixer::processRecordingMissingPackets(const uint64_t timestamp)
+void EngineMixer::processRecordingUnackedPackets(const uint64_t timestamp)
```
olofkallander, line 3649: What this method actually does is to resend cached packets if they are not acked yet.
The shouldProcess is data race when called from here but we have not seen it in tcheck yet because we do not test recording.

bridge/engine/ProcessMissingVideoPacketsJob.cpp
```diff
@@ -27,14 +27,14 @@ void ProcessMissingVideoPacketsJob::run()
 {
     auto timestamp = utils::Time::getAbsoluteTime();
     auto* videoMissingPacketsTracker = _ssrcContext.videoMissingPacketsTracker.get();
-    if (!videoMissingPacketsTracker || !videoMissingPacketsTracker->shouldProcess(timestamp / 1000000ULL))
+    if (!videoMissingPacketsTracker || !videoMissingPacketsTracker->shouldProcess(timestamp))
```
olofkallander, line 30: I prefer if we are consistently using the absolute timestamp. Also these divisions will fail if the timestamp would start at a point that wraps soon.

bridge/engine/UnackedPacketsTracker.h
```diff
@@ -161,6 +154,7 @@ class UnackedPacketsTracker
             _unackedPackets.erase(entriesToErase[i]);
         }
 
+        _lastRunTimestampMs = timestampMs;
```
olofkallander, line 157: This tracker should also convert to absolute timestamp.

bridge/engine/ProcessUnackedRecordingEventPacketsJob.cpp
```diff
@@ -26,9 +26,15 @@ ProcessUnackedRecordingEventPacketsJob::ProcessUnackedRecordingEventPacketsJob(
 
 void ProcessUnackedRecordingEventPacketsJob::run()
 {
+    uint64_t timestamp = utils::Time::getAbsoluteTime();
+    if (!_unackedPacketsTracker.shouldProcess(timestamp / utils::Time::ms))
```
olofkallander, line 30: This solves data race with Engine thread

RicardoMDomingues: It's a bit unfortunate that we have to create jobs that have nothing to do. This is packets have a very low frequency. In some recording we can have long periods (like minutes) without a single packet. It would be nice if we had a fast check in engine thread to not run this if there is zero packets to track

olofkallander: I agree. For the video rtx we want to request packets asap, so we have them before the client requests them also.
I guess the only way to signal this is to have an atomic flag set by transport when there is a potential loss detected / need to request. Then that flag will be used to trigger the job. On the other hand, atomic flags have a cost too.

bridge/engine/EngineMixer.cpp
```diff
@@ -3646,18 +3648,20 @@ void EngineMixer::processBarbellMissingPackets(bridge::SsrcInboundContext& ssrcI
...
     for (auto& engineRecordingStreamEntry : _engineRecordingStreams)
     {
         auto engineRecordingStream = engineRecordingStreamEntry.second;
         for (auto& recEventMissingPacketsTrackerEntry : engineRecordingStream->recEventUnackedPacketsTracker)
         {
             auto& recEventMissingPacketsTracker = recEventMissingPacketsTrackerEntry.second;
-            if (!recEventMissingPacketsTracker.shouldProcess(timestamp / 1000000ULL))
-            {
-                continue;
-            }
```
RicardoMDomingues, line 3660: I guess we could have a check like this
```
if (!recEventMissingPacketsTracker.empty())
            {
                continue;
            }
```

RicardoMDomingues (approved):

PR #196, opened 2022-09-23
activeMedia flag bug

olofkallander:


bridge/engine/EngineMixer.cpp
```diff
@@ -3235,7 +3236,10 @@ void EngineMixer::sendDominantSpeakerToRecordingStream(EngineRecordingStream& re
     }
 }
 
-void EngineMixer::updateSimulcastLevelActiveState(EngineVideoStream& videoStream,
+// This method is called after a video stream has been reconfigured. That has lead to streams being removed and then
```
reddvl1980, line 3239: Typo in "has lead" -> "has led".

reddvl1980 (approved):

PR #198, opened 2022-09-28
Fix broken deletion check

marcusspangenberg:
Outbound contexts should be deleted if they are markedForDeletion and become idle, not as soon as they become idle.

What happens now is that outboundContexts will be deleted when they are idle, even if they will resume later on. The SRTP context is then deleted and the receiving end will get SRTP unprotect errors.

This bug was introduced in PR #150

### DIFF
--- a/bridge/VideoStream.h
+++ b/bridge/VideoStream.h
@@ -35,10 +35,10 @@ struct VideoStream
     {
     }
 
-    std::string id;
-    std::string endpointId;
-    size_t endpointIdHash;
-    uint32_t localSsrc;
+    const std::string id;
+    const std::string endpointId;
+    const size_t endpointIdHash;
+    const uint32_t localSsrc;
     SimulcastStream simulcastStream;
     utils::Optional<SimulcastStream> secondarySimulcastStream;
     std::shared_ptr<transport::RtcTransport> transport;
@@ -50,7 +50,7 @@ struct VideoStream
 
     bool markedForDeletion;
     const bool ssrcRewrite;
-    bool isDtlsLocalEnabled;
+    const bool isDtlsLocalEnabled;
     bool isConfigured;
     const uint32_t idleTimeoutSeconds;
 };
--- a/bridge/VideoStream.h
+++ b/bridge/VideoStream.h
@@ -49,7 +49,7 @@ struct VideoStream
     SsrcWhitelist ssrcWhitelist;
 
     bool markedForDeletion;
-    bool ssrcRewrite;
+    const bool ssrcRewrite;
     bool isDtlsLocalEnabled;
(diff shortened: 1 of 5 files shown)


olofkallander:
I have not completed all review, but from the description above, isnät the problem that the srtp state is removed when stream turns idle and is then resumed. We do not initiate deletion of the entire outbound context when it turns idle. If the deletion flag is set the context is already about to be eleted becuase the entire streams was removed. In that case there is no point in doing idle processing beacuse everything will be deleted.
I agree that there is a bug but I am not sure it is the markedForDeletion flag that is the fix.

marcusspangenberg:
> I have not completed all review, but from the description above, isnät the problem that the srtp state is removed when stream turns idle and is then resumed. We do not initiate deletion of the entire outbound context when it turns idle. If the deletion flag is set the context is already about to be eleted becuase the entire streams was removed. In that case there is no point in doing idle processing beacuse everything will be deleted. I agree that there is a bug but I am not sure it is the markedForDeletion flag that is the fix.

The intent of the logic was that checkPacketCounters deletes idle contexts only if they are also markedForDeletion. markedForDeletion is set on all related outbound contexts when a video stream is removed. Therefore, checkPacketCounters should tag a context as idle when not markedForDeletion, but not delete it or delete the srtp context. If checkPacketCounters detects an idle context that is also markedForDeletion, it should delete it. It used to work like that at least.

olofkallander:
yes, I have walked through the code to get the original intent and how mark and idle work together. 
So one thing I will do for sure is to not delete any outbound contexts for participants that support ssrc rewrite. It is not useful at all.
So if anything should be cleaned up, it is only for clients not supporting rewrite. If so it should be synchorinezed with jobs over transport so outbound context is removed as soon as previous send jobs have completed and the ssrc has been removed from srtp.
The other easy alternative is to just leave outbound context in there as those clients should be few and all should support rewrite. In new api it is not even an option to join without rewrite I think.

marcusspangenberg:
> yes, I have walked through the code to get the original intent and how mark and idle work together. So one thing I will do for sure is to not delete any outbound contexts for participants that support ssrc rewrite. It is not useful at all. So if anything should be cleaned up, it is only for clients not supporting rewrite. If so it should be synchorinezed with jobs over transport so outbound context is removed as soon as previous send jobs have completed and the ssrc has been removed from srtp. The other easy alternative is to just leave outbound context in there as those clients should be few and all should support rewrite. In new api it is not even an option to join without rewrite I think.

Great! Thanks for taking the time.

RicardoMDomingues (approved):

marcusspangenberg:
@olofkallander ready to merge?

PR #199, opened 2022-09-28
Fix corrupt RTCP

RicardoMDomingues:


transport/TransportImpl.cpp
```diff
@@ -1691,10 +1694,19 @@ void TransportImpl::sendReports(uint64_t timestamp, bool rembReady)
             rembAdded = true;
         }
 
-        if (rtcpPacket->getLength() + MINIMUM_SR + std::min(receiverReportCount, 4u) * sizeof(rtp::ReportBlock) >
-            _config.mtu)
+        assert(rtcpPacket->getLength() <= packetLimit);
+
+        const auto spaceAvailable = packetLimit - rtcpPacket->getLength();
```
olofkallander, line 1699: We cannot use packetLimit as limit. Remove that from Packet.h and use _config.mtu instead. Otherwise the packet may grow to large to be sent and it is dropped.

transport/TransportImpl.cpp
```diff
@@ -1648,6 +1648,9 @@ void TransportImpl::sendReports(uint64_t timestamp, bool rembReady)
     bool rembAdded = false;
     memory::UniquePacket rtcpPacket;
     uint32_t senderSsrc = 0;
+    const size_t packetLimit = std::min(static_cast<size_t>(_config.mtu), memory::Packet::maxLength());
```
olofkallander, line 1651: ok I see you put limit here from MTU.

transport/TransportImpl.cpp
```diff
@@ -1648,6 +1648,9 @@ void TransportImpl::sendReports(uint64_t timestamp, bool rembReady)
...
+    const size_t maxReceiveReportsBySenderReport = 15;
```
olofkallander, line 1652: Should we go with MAX_RR_COUNT or MAX_RECEIVE_REPORT_COUNT ?

transport/TransportImpl.cpp
```diff
@@ -1648,6 +1648,9 @@ void TransportImpl::sendReports(uint64_t timestamp, bool rembReady)
...
+    size_t nextMaxReportBlocks = maxReceiveReportsBySenderReport;
```
olofkallander, line 1653: reportBlocksToAdd
and then do --reportBlocksToAdd instead of use k

transport/TransportImpl.cpp
```diff
@@ -1691,10 +1694,19 @@ void TransportImpl::sendReports(uint64_t timestamp, bool rembReady)
             rembAdded = true;
```
olofkallander, line 1694: This is perhaps extending the task, but could we assure the REMB is actually added to the first report packet. It is not big but it is sad if it is not added until the last report. It is information that far side needs asap.
Something like adding one RB less to the frist SR / RR if it allows REMB to be added. I notice in the appendreMb that it just returns if it does not fit and we think it has been added.

RicardoMDomingues: It will be send always on 1st SR/RR as the max repots block we add is 15, then there is always space for REMB on the 1st SS/RR (at least with 1480 mtu configuration). But I will ensure in the case for some reason mtu is configured to a very low number

olofkallander: super I see that. A packet MTU will at least be  1200 so REMB should fit right after first SR even if it is full of RBs.

transport/RtcpReportsProducer.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+    static constexpr uint64_t ntp32Tick = 0x10000u; // 1/65536 sec
+    const size_t packetLimit = std::min(static_cast<size_t>(_config.mtu), memory::Packet::maxLength());
+    bool rembSent = false;
+    size_t rembSize = reportContext.rembPacket.getLength();
+
+    uint32_t nextReportBlocksCount = std::min(reportContext.receiverReportCount, rtp::RtcpHeader::MAX_REPORT_BLOCKS);
+    if (rembSize > 0)
+    {
+        // The assumption is this is the 1st method called then rtcpPacket is still is a null pointer
```
olofkallander, line 147: remove last "is"

utils/Span.h
```diff
@@ -0,0 +1,90 @@
...
+{
+
+/**
+ * The class template span describes an object that can refer to a contiguous sequence of objects with the first element
+ * of the sequence at position zero.
+ *
+ * This is intends to be an implementation of std::span as it only available on C++20.
+ */
+template <class T>
+class Span
```
olofkallander, line 15: Quite useful to turn raw vectors into iteratable container. We probably have several places where we can apply it in iterations.

transport/RtcpReportsProducer.cpp
```diff
@@ -0,0 +1,335 @@
...
+    uint32_t senderReportSsrcs[_outboundSsrcCounters.capacity()];
+    uint32_t receiverReportSsrcs[_inboundSsrcCounters.capacity()];
+    uint32_t activeSsrcs[_inboundSsrcCounters.capacity()];
+
+    ReportContext reportContext;
+    reportContext.senderReportSsrcs = utils::Span<uint32_t>(senderReportSsrcs, _outboundSsrcCounters.capacity());
+    reportContext.receiverReportSsrcs = utils::Span<uint32_t>(receiverReportSsrcs, _inboundSsrcCounters.capacity());
+    reportContext.activeSsrcs = utils::Span<uint32_t>(activeSsrcs, _inboundSsrcCounters.capacity());
+
+    collectSsrcsReadyToSendSenderReports(reportContext, timestamp);
```
olofkallander, line 52: inside this method. Why do we track sr count in separate value in ReportContext. Would it not be cleaner to update the Span instead?

RicardoMDomingues: Yes, I will do that

utils/Span.h
RicardoMDomingues: Yes and I see other situations where this can be useful too. Like when we want want give just a view of a subpart a vector/array. And to have a similar interface regardless what the container is so we can use utils::Span instead of using templates to receive parameters that can be Carrays, std::arrays or std::vectors

rtp/RtcpHeader.cpp
```diff
@@ -399,7 +401,7 @@ RtcpSenderReport::RtcpSenderReport() : ssrc(0)
 RtcpSenderReport* RtcpSenderReport::create(void* buffer)
 {
     auto report = reinterpret_cast<RtcpSenderReport*>(buffer);
-    std::memset(report, 0, sizeof(*report));
+    std::memset(report, 0, RtcpSenderReport::minimumSize());
```
RicardoMDomingues, line 404: The real problem was in here

test/transport/RtcpReportsProducerTest.cpp
```diff
@@ -0,0 +1,676 @@
...
+#include "rtp/RtpHeader.h"
+#include "transport/RtcpReportProducer.h"
+#include <gmock/gmock.h>
+
+using namespace transport;
+using namespace testing;
+
+namespace
+{
+const uint64_t interval = 600000000;
```
olofkallander, line 13: better to use ' or *utils::ms

rtp/RtcpHeader.h
```diff
@@ -29,6 +29,8 @@ struct RtcpHeader
     uint8_t packetType : 8;
     nwuint16_t length;
 
+    static constexpr size_t maxReportsBlocks() { return 31; }
```
olofkallander, line 32: What was the reason to do these static const methods instead of a public static const member that can read like a property? I think you mentioned it that it was related to template use.

transport/RtcpReportProducer.h
```diff
@@ -0,0 +1,59 @@
...
+        virtual ~RtcpSender() = default;
+        virtual void sendRtcp(memory::UniquePacket packet, uint64_t timestamp) = 0;
+    };
+
+    RtcpReportProducer(const logger::LoggableId& loggableId,
+        const config::Config& config,
+        const concurrency::MpmcHashmap32<uint32_t, RtpSenderState>& outboundSsrcCounters,
+        const concurrency::MpmcHashmap32<uint32_t, RtpReceiveState>& inboundSsrcCounters,
+        memory::PacketPoolAllocator& rtcpPacketAllocator,
+        RtcpSender& rtcpSender);
```
olofkallander, line 29: I would like to avoid these callback loops when possible. In several other similar situations we ask the "factory" repeatedly for data until it returns false or null. The factory can maintain state easily in most cases.

It makes it a little bit more decoupled and easier to follow call stacks when they are more of a tree than a graph.

transport/TransportImpl.h
```diff
@@ -50,7 +51,8 @@ class TransportImpl : public RtcTransport,
                       public sctp::DatagramTransport,
                       public sctp::SctpServerPort::IEvents,
                       public sctp::SctpAssociation::IEvents,
-                      public ServerEndpoint::IEvents
+                      public ServerEndpoint::IEvents,
+                      private RtcpReportProducer::RtcpSender
```
olofkallander, line 55: See comment on RtcpSender

rtp/RtcpHeader.h
RicardoMDomingues: On c++14 inline static variables are supported. They worked only if you use them inside of the scope where they are declared. But if you try to use it outside the linker will fail because by the linkage entry does not exists.

We have 2 option or use a static variable but assign the value on cpp file (but it will not be a constexpr anylonger) or use a constexpr function

transport/RtcpReportProducer.h
RicardoMDomingues: I can do that but I think it will cause more complexity in this case. Because we have to keep the state of ReportContext between multiple calls to generate the report. Moreover, we use stack memory to store the ssrc to send. That change would require a big change in the implementation of RtcReportProducer

test/transport/RtcpReportsProducerTest.cpp
RicardoMDomingues: This should have same value of configuration. Then deleted this line and I am reading the value directly from `_config` instead

transport/RtcpReportProducer.cpp
```diff
@@ -0,0 +1,380 @@
...
+    if (!packet)
+    {
+        packet = memory::makeUniquePacket(allocator);
+    }
+
+    return !!packet;
+}
+
+// @return the removed tail span
+utils::Span<uint32_t> truncateTail(utils::Span<uint32_t>& ssrcReport, size_t n)
```
olofkallander, line 22: I wonder if this should be a helper method in Span.h

olofkallander (approved):

transport/RtcpReportProducer.cpp
RicardoMDomingues: I thought about that too. But then I thought it would be easier to replace `utils::span` to `std::span` if we implement exactly the same methods (if one day we update C++ version to C++20 of curse)

PR #200, opened 2022-09-29
MixerManager task queues

olofkallander:
demo to discuss how long run jobs could be handled, and perhaps specialized message queues avoided.


bridge/engine/EngineMessageListener.h
```diff
@@ -1,16 +1,22 @@
...
+class EngineMixer;
 
 class EngineMessageListener
 {
 public:
     virtual ~EngineMessageListener() = default;
 
     virtual bool onMessage(EngineMessage::Message&& message) = 0;
+
+    virtual std::shared_ptr<Mixer> onEngineMixerRemoved1(EngineMixer& mixer) = 0;
```
olofkallander, line 18: This is a bit of hack. We do not really want to abuse this interface like this.

bridge/MixerManager.cpp
```diff
@@ -365,51 +394,35 @@ bool MixerManager::onMessage(EngineMessage::Message&& message)
...
+        return mixer;
+    }
 
-            _mixers.erase(findResult);
-        }
-        else
-        {
-            logger::debug("did not find mixer to stop %s", "MixerManager", mixerId.c_str());
-        }
+    return std::shared_ptr<Mixer>();
```
RicardoMDomingues, line 411:
```suggestion
    return nullptr;
```

easer to read

bridge/engine/EngineMessageListener.h
RicardoMDomingues: I think this methods shouldn't be part of `EngineMessageListener`.`EngineMixerRemoved` has direct access to `MixerManager` type.
Methods name with "1" and "2" are also not great.

bridge/engine/Engine.h
```diff
@@ -50,6 +55,7 @@ class Engine
     uint32_t _tickCounter;
 
     concurrency::MpmcQueue<utils::Function> _threadQueue;
```
olofkallander, line 57: If this is a JobManager instead, we can let the Engine thread pick jobs from it and we can still post Callable jobs on it that allows direct method invocation.

bridge/Mixer.cpp
```diff
@@ -1390,29 +1343,25 @@ bool Mixer::reconfigureVideoStream(const std::string& endpointId,
...
-        command.command.reconfigureVideoStreamSecondary.simulcastStream = videoStream->simulcastStream;
-        command.command.reconfigureVideoStreamSecondary.secondarySimulcastStream =
-            videoStream->secondarySimulcastStream.get();
-        command.command.reconfigureVideoStreamSecondary.transport = videoStream->transport.get();
-        std::memcpy(&command.command.reconfigureVideoStreamSecondary.ssrcWhitelist,
-            &ssrcWhitelist,
-            sizeof(SsrcWhitelist));
+        _engine.post(utils::bind(&EngineMixer::reconfigureVideoStream,
+            _engineMixer.get(),
+            videoStream->transport.get(),
```
olofkallander, line 1350: What if we have another reconfigure before the engine thread completes this. We will have data in videoStream that is being updated.

RicardoMDomingues: Perhaps it would be better give a copy of `ssrcWhitelist` and simulcasts if that fits in `utils::function` storage

bridge/MixerManagerAsync.h
```diff
@@ -0,0 +1,56 @@
...
+    virtual void videoStreamRemoved(EngineMixer* engineMixer, EngineVideoStream* videoStream) = 0;
+    virtual void sctpReceived(EngineMixer* mixer, memory::UniquePacket msgPacket, size_t endpointIdHash) = 0;
+    virtual void dataStreamRemoved(EngineMixer* mixer, EngineDataStream* dataStream) = 0;
+    virtual void freeRecordingRtpPacketCache(EngineMixer* mixer, uint32_t ssrc, size_t endpointIdHash) = 0;
+    virtual void barbellRemoved(EngineMixer* mixer, EngineBarbell* barbell) = 0;
+    virtual void recordingStreamRemoved(EngineMixer* mixer, EngineRecordingStream* recordingStream) = 0;
+    virtual void removeRecordingTransport(EngineMixer* mixer, const char* streamId, size_t endpointIdHash) = 0;
+    virtual void inboundSsrcContextRemoved(EngineMixer* mixer, uint32_t ssrc, codec::OpusDecoder* opusDecoder) = 0;
+    virtual void mixerTimedOut(EngineMixer* mixer) = 0;
+    virtual void engineRecordingStopped(EngineMixer* mixer, RecordingDescription* recordingDesc) = 0;
```
RicardoMDomingues, line 53: Isn't better to receive references instead of pointers that we know that can't be null? To use references with `utils::bind`, you can use `std::ref` and `std::cref`

bridge/engine/SsrcWhitelist.h
```diff
@@ -8,6 +8,15 @@ namespace bridge
 
 struct SsrcWhitelist
 {
+    SsrcWhitelist() : enabled(false), numSsrcs(0), ssrcs{0} {}
+    SsrcWhitelist& operator=(const SsrcWhitelist& o)
```
RicardoMDomingues, line 12: I guess we shouldn't specify a copy/move assign/ctor when it is not change the default behaviour. Besides being unnecessary code, this class has all requirements to be trivially copiable but this explicit copy assignment make it non-trivial and some compiler's optimisations opportunities are lost

olofkallander:
> I guess we shouldn't specify a copy/move assign/ctor when it does not change the default behaviour. Besides being unnecessary code, this class has all requirements to be trivially copiable but this explicit copy assign implementation make it non-trivial and some compiler's optimisations opportunities can be lost 

I assumed that too, but it would not accept it without explicit assign operator. I can try again and search for root cause.

RicardoMDomingues (approved):

PR #210, opened 2022-10-12
fix crash

olofkallander:


RicardoMDomingues (approved):

bridge/engine/EngineMixer.cpp
```diff
@@ -4014,143 +4061,148 @@ void EngineMixer::onBarbellUserMediaMap(size_t barbellIdHash, const char* messag
...
-        fwdVideoEndpoints.add(item.endpointIdHash, item.endpointIdHash);
+        if (stream.endpointIdHash.isSet())
+        {
+            addToMap(stream, videoSsrcs);
+        }
     }
-
-    memory::Map<size_t, size_t, 32> fwdAudioEndpoints;
-    for (auto& item : audioSsrcs)
+    if (barbell->slideStream.endpointIdHash.isSet())
```
RicardoMDomingues, line 4073: I think store `slideStream` separated of `videoStreams` gave us nothing than confusion. We are used to handle slideStreams as part of videoStreams for the real users and this separation already cause some problems by forgetting to process this stream.

Wouldn't it be better add it to videoStreams instead?

bridge/engine/EngineMixer.cpp
```diff
@@ -4014,143 +4061,148 @@ void EngineMixer::onBarbellUserMediaMap(size_t barbellIdHash, const char* messag
...
+                }
+                videoStream->endpointIdHash.set(entry.first);
+                videoStream->endpointId.set(item.endpointId);
             }
-            else if (audioStream->endpointIdHash.isSet() && audioStream->endpointIdHash.get() != item.endpointIdHash)
+
+            _activeMediaList->addBarbellVideoParticipant(entry.first, primary, secondary, item.endpointId);
+            _engineStreamDirector->addParticipant(entry.first, primary, secondary.isSet() ? &secondary.get() : nullptr);
+
+            for (const auto& simLevel : utils::Span<SimulcastLevel>(primary.levels, primary.numLevels))
```
RicardoMDomingues, line 4144: I think this will became a regular pattern. Maybe it would be better to create a method on `SimulcastStream`  to return `utils::Span<SimulcastLevel>` (example,  `primary.getActiveLevels()`)

bridge/engine/EngineMixer.cpp
olofkallander: The reasons was to be able to identify which stream is slides without searching as it is special and may not be used by video. But it seems we do not use that yet.

PR #212, opened 2022-10-18
pin BB fix

olofkallander:


reddvl1980:
LGTM

reddvl1980 (approved):
@olofkallander , please check the commit I've added regading keeping local var "const auto".

bridge/engine/EngineStreamDirector.h
```diff
@@ -697,7 +703,7 @@ class EngineStreamDirector
         const auto participantStreamsItr = _participantStreams.find(endpointIdHash);
         if (participantStreamsItr == _participantStreams.end())
         {
-            return lowQuality;
+            return dropQuality;
```
olofkallander, line 706: wdyt about this one. It seems reasonable to me but perhaps there is something I missed on how Director works.

PR #213, opened 2022-10-18
RTC-13306: Ensure safe deletion of OutboundSsrcContexts

RicardoMDomingues:


bridge/engine/EngineFunction.h
```diff
@@ -0,0 +1,249 @@
...
+    TThis* copyTo(EngineFunctionStorage& newStorage) const final { return new (&newStorage) TThis(*this); }
+
+private:
+    Func mEngineFunction;
+    std::tuple<Args...> mFunctionArguments;
+};
+
+} // namespace detail
+
+class EngineFunction
```
RicardoMDomingues, line 118: Engine function is a sort of of std::function with a forced inlined storage  that never allocate dynamic memory so we can use it safely on EngineThread

test/bridge/EngineFunctionTest.cpp
```diff
@@ -0,0 +1,290 @@
...
+    StrictMock<CallbackClassMock> callbacks6;
+    StrictMock<CallbackClassMock> callbacks7;
+
+    auto callbackBind0 = bridge::engine::bind(&CallbackClassMock::callback0, &callbacks0);
+    auto callbackBind1 = bridge::engine::bind(&CallbackClassMock::callback1, &callbacks1, 5);
+    auto callbackBind2 = bridge::engine::bind(&CallbackClassMock::callback2, &callbacks2, 10, 20);
+    auto callbackBind3 = bridge::engine::bind(&CallbackClassMock::callback3, &callbacks3, 30, 40, true);
+    auto callbackBind4 = bridge::engine::bind(&CallbackClassMock::callback4, &callbacks4, 30, 40, false, fakeValue0);
+    auto callbackBind5 =
+        bridge::engine::bind(&CallbackClassMock::callback5, &callbacks5, 30, 40, true, fakeValue0, fakeValue1);
```
RicardoMDomingues, line 112: Here is an example how we can use `EngineFunction` using `engine::bind`

bridge/engine/EngineMixer.cpp
```diff
@@ -179,6 +164,61 @@ class RemovePacketCacheJob : public jobmanager::CountedJob
...
+
+        _transport.removeSrtpLocalSsrc(ssrc);
+
+        _engineThreadContext.post(engine::bind(&EngineMixer::onOutboundContextFinalized,
+            &_mixer,
+            endpointIdHash,
+            ssrc,
+            _feedbackSsrc,
+            _outboundContext.rtpMap.isVideo()));
+    }
```
RicardoMDomingues, line 211: This task finalize for non ssrc-rewrite outbound contexts. It must be called when the associated inbound stream is removed. Will do all this in a single Job:

- Send Rtcp Goodbye signalling that ssrc will be removed
- Delete video packet cache when outbound ctx keeps one
- Remove SrtpSsrc (does what old RemoveSrtpSsrcJob did)
- Call the EngineMixer on EngineThread to the context to be removed. This MUST be the last job run for this OuboundContext

bridge/engine/EngineMixer.cpp
```diff
@@ -1839,14 +1886,40 @@ void EngineMixer::onRtcpPacketDecoded(transport::RtcTransport* sender,
...
     {
+        if (ssrcOutboundContextItr->second.markedForDeletion)
+        {
+            // The outboundContext is marked for deletion after owner inbound stream
+            // as been removed (only for non ssrcRewrite). This means the job to
+            // remove OutboundSsrc has already been enqueued and it will be deleted
+            // soon. Then we MUST NOT enqueue any more jobs with a reference to this
+            // SsrcOutboundContext
+            assert(!ssrcRewrite);
+            return nullptr;
```
RicardoMDomingues, line 1903: If it's marked for deletion is because `FinalizeNonSsrcRewriteOutboundContextJob` has been already enqueue. No more jobs can use this context. After `FinalizeNonSsrcRewriteOutboundContextJob` has run successfully, it will post an execution on EngineThread to delete the SsrcOuboundContext

bridge/engine/EngineMixer.cpp
```diff
@@ -1839,14 +1886,40 @@ void EngineMixer::onRtcpPacketDecoded(transport::RtcTransport* sender,
...
+    {
+        // When is not ssrc rewrite, we have to ensure that ssrcInboundContext is not been
+        // decommissioned already. If it was decommissioned and the ssrcOutboundContext
+        // is not in the ssrcOutboundContexts, most likely the reason is because we are
+        // processing an old packet in the queue and the ssrcOutboundContext has been deleted already.
+        // In such case we we will not recreate the outboundContext because it would be
+        // deleted only in the end of the conference (keep with it all associated resources, like video caches)
+        if (!_ssrcInboundContexts.contains(ssrc))
+        {
+            return nullptr;
```
RicardoMDomingues, line 1919: I think this never happens, but f the SsrcInboundContext does not exist (means it was decommissed) is because we are processing a packet for a stream that was removed recently so we will not create an outboundContext because most likely it was deleted recently also. Moreover, if inbound stream was removed and we create an OutboundContext again, it would be removed only when the oubound stream was removed.

This should never hit here when SsrcInboundContext id decommissed by timeout. Otherwise, it would mean that packet is more than 5 minutes old

bridge/engine/EngineMixer.cpp
```diff
@@ -452,6 +491,17 @@ void EngineMixer::removeStream(EngineVideoStream* engineVideoStream)
         }
     }
 
+    for (auto& ssrcOutboundContextPair : engineVideoStream->ssrcOutboundContexts)
+    {
+        if (!ssrcOutboundContextPair.second.markedForDeletion)
+        {
+            engineVideoStream->transport.getJobQueue().addJob<RemovePacketCacheJob>(*this,
```
RicardoMDomingues, line 498: When we remove video stream we have to remove packet cache as well, otherwise it will be removed only when the Mixer is deleted. This because `_videoPacketCaches` in `Mixer`  is deleted when message `FreeVideoPacketCache` is received only

bridge/engine/EngineThreadContext.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "EngineFunction.h"
+#include "concurrency/MpmcQueue.h"
+
+namespace bridge
+{
+
+class EngineThreadContext
```
RicardoMDomingues, line 9: Easy way to WT to "post" task to EngineThread without create a round trip of specific message to MixerManager and MixerManager to Engine

bridge/engine/EngineFunction.h
```diff
@@ -0,0 +1,249 @@
...
+{
+    return !!f;
+}
+
+namespace engine
+{
+
+template <class Func, class... Args>
+std::enable_if_t<std::is_function<std::remove_pointer_t<Func>>::value, detail::EngineFunctionBinder<Func, Args...>>
+bind(Func&& func, Args&&... args)
```
RicardoMDomingues, line 232: Use in same way of `std::bind` but without placeholder suport

bridge/engine/EngineFunction.h
```diff
@@ -0,0 +1,249 @@
...
+{
+    return detail::EngineFunctionBinder<Func, std::decay_t<Args>...>(std::forward<Func>(func),
+        std::forward<Args>(args)...);
+}
+
+template <class T, class U, class K, class... Args>
+auto bind(U T::*memberFunction, K* instance, Args&&... args)
```
RicardoMDomingues, line 239: Use in same way of std::bind but without placeholder suport

bridge/engine/EngineFunction.h
```diff
@@ -0,0 +1,254 @@
...
+
+private:
+    friend class ::bridge::EngineFunction;
+    virtual void invoke() const = 0;
+    virtual Invokable* moveTo(EngineFunctionStorage& newStorage) = 0;
+    virtual Invokable* copyTo(EngineFunctionStorage& newStorage) const = 0;
+};
+
+template <class Func, class... Args>
+class EngineFunctionBinder final : public Invokable
```
olofkallander, line 41: Should we not rename this and move it outside bridge? It is generically reusable and the EngineFunction could be defined as a type of the GenericFunction (will ponder a bit more on the name) and specification of the storage size.

bridge/engine/EngineThreadContext.h
olofkallander: Here I think we should call it what it is, a sort of TaskQueue or something.

olofkallander: Maybe not in this same PR, but we should consider using this in MixerManager and Engine if we think this is the way to handle pending tasks / jobs. 
We have 3 methods now:
- POD type union messages
- Job based class per method call
- Task / serialized function call queue
We should perhaps replace the POD type one. This could have been solved by posting a POD type onto the engine command queue afaik, but it would have disadvantages like limitations on the objects passed in arguments.

bridge/engine/Engine.h
```diff
@@ -16,7 +17,7 @@ namespace bridge
 class EngineMixer;
 class EngineMessageListener;
 
-class Engine
+class Engine : private EngineThreadContext
```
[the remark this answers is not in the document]
RicardoMDomingues: I didn't use the composition to not expose `processEngineTasks`. By using a private inheritance we can have `processEngineTasks` `protected` and it will not be visible that `Engine` is `EngineThreadContext ` for outside.

Use composition I guess we need to split `EngineThreadContext` in two class with inheritance also. On top we will have only the post methods. And on the button the methods that will process the queue so the type you give to outside only have access to `post` methods

bridge/engine/EngineMixer.cpp
```diff
@@ -1202,27 +1295,59 @@ void EngineMixer::checkInboundPacketCounters(const uint64_t timestamp)
...
 
-                EngineMessage::Message message(EngineMessage::Type::FreeRecordingRtpPacketCache);
-                message.command.freeRecordingRtpPacketCache.mixer = this;
-                message.command.freeRecordingRtpPacketCache.ssrc = outboundContext.ssrc;
-                message.command.freeRecordingRtpPacketCache.endpointIdHash = recordingStreamEntry.first;
-                _messageListener.onMessage(std::move(message));
+                const auto ssrc = outboundContextEntry.first;
+                outboundContext.markedForDeletion = true;
+                auto& outboundContextsFinalizerCounter = recordingStreamEntry.second->outboundContextsFinalizerCounter;
+                auto counterIt = outboundContextsFinalizerCounter.find(ssrc);
```
RicardoMDomingues, line 1323: This is a quite complex, EngineRecordingStream can have multiple transports (redundancy feature that we don't use yet and don't know if we will use it). But that create complex on syncronization has we need to ensure that all transports have executed the finalised before remove the outbound context.

outboundContextsFinalizerCounter is a stores a atomic counter of finalizers to know we have run all finalize jobs before remove the context

bridge/engine/EngineFunction.h
```diff
@@ -0,0 +1,254 @@
...
+
+class EngineFunction;
+
+namespace detail
+{
+
+constexpr size_t calculateStorageSize()
+{
+    constexpr size_t minSize = 128;
+    // EngineFunction will have EngineFunctionStorage + a pointer
```
olofkallander, line 15: I am very much against these long explanations in between the lines of code. If really necessary, put them above the method and try to condense it. I am sure we can explain the same thing in half the number of words.

bridge/engine/EngineMixer.cpp
```diff
@@ -1839,14 +1980,38 @@ void EngineMixer::onRtcpPacketDecoded(transport::RtcTransport* sender,
...
-    const RtpMap& rtpMap)
+    const RtpMap& rtpMap,
+    const bool ssrcRewrite)
 {
     auto ssrcOutboundContextItr = ssrcOutboundContexts.find(ssrc);
     if (ssrcOutboundContextItr != ssrcOutboundContexts.cend())
     {
+        if (ssrcOutboundContextItr->second.markedForDeletion)
+        {
+            // The outboundContext is marked for deletion after owner inbound stream
```
olofkallander, line 1991: I can see in code it is marked for deletion. Hopefully it is clear from other code that the remove job has been posted, but if we want to mention it reduntantly here the comment could be.
"RemoveOutboundSsrc job posted.
Do not use this outbound context."

But even that is evident from what the code does, return nunllptr and exit here.

bridge/engine/EngineThreadContext.h
olofkallander: What is actually the problem of making a generic Job that can take on that "on stack" Function type?
I assume there are some catches in the details.

olofkallander: Then we would have one JobQueue to pull from instead of two different ones.

bridge/engine/SsrcOutboundContext.h
```diff
@@ -104,6 +105,11 @@ class SsrcOutboundContext
 
     // Stream owner is being removed. Stop outbound packets over this context
     bool markedForDeletion;
+
+    // This is for recording only! We will keep the SsrcOutbound context for while
```
olofkallander, line 109: I see it is for recording by the variable name. The question is if the rest is bringing meaning to the reader.

// Retain rec OutboundSsrc before marking for deletion to sustain retransmissions longer.

bridge/engine/EngineFunction.h
olofkallander: utils::Function maybe would work.

bridge/engine/EngineThreadContext.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+size_t EngineThreadContext::processEngineTasks(size_t maxTasksToProcess)
+{
+    size_t processedTasks = 0;
+    const size_t tasksToProcess = std::min(_threadTasksQueue.size(), maxTasksToProcess);
+
+    for (size_t i = 0; i < tasksToProcess; ++i)
+    {
+        utils::Function task;
+        // We could specialize thread task queue so we can execute
```
RicardoMDomingues, line 13: Remove the comment

jobmanager/Job.h
```diff
@@ -48,4 +48,38 @@ class CountedJob : public Job
...
+    explicit CallableJob(const Callable& callable) : _callable(callable) {}
+    explicit CallableJob(Callable&& callable) : _callable(std::move(callable)) {}
+
+    void run() final { _callable(); }
+
+private:
+    Callable _callable;
+};
+
+template <class Callable>
```
RicardoMDomingues, line 64: Generic callable. The only restriction is that needs to implement `operator()()`. This is has some advantages over give a `std::function.`

1. We can use the direct type returned by `utils::bind` which creates one less level of indirection as the result is a `final` type and does not need to invoke more virtual methods
2. We can use lambda functions directly. The lambda function itself does not created dynamic memory (dynamic memory can happens when we transform the lambda on std::function).

transport/TransportImpl.cpp
```diff
@@ -1050,7 +993,14 @@ void TransportImpl::onIceTcpConnect(std::shared_ptr<Endpoint> endpoint,
 {
     if (_rtpIceSession && _isRunning)
     {
-        if (!_jobQueue.addJob<TcpConnected>(*this, endpoint, source, std::move(packet)))
+        const bool posted = _jobQueue.post(_jobCounter,
+            utils::bind(&TransportImpl::internalIceTcpConnect,
+                this,
+                std::cref(endpoint),
+                std::cref(target),
+                utils::rvalueParam(std::move(packet))));
```
RicardoMDomingues, line 1001: Need use `utils::rvalueParam` to pass packets are copy constructor is deleted

transport/TransportImpl.cpp
```diff
@@ -2150,7 +2102,7 @@ bool TransportImpl::sendSctp(const uint16_t streamId,
 
 void TransportImpl::connectSctp()
 {
-    _jobQueue.addJob<ConnectSctpJob>(*this);
+    _jobQueue.post(_jobCounter, [this]() { doConnectSctp(); });
```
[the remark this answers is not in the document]
RicardoMDomingues: Example of callstack for transport conect
```
transport::TransportImpl::doConnect() 
transport::TransportImpl::connect()::$_1::operator()() const 
jobmanager::CallableCountedJob<transport::TransportImpl::connect()::$_1>::run()
jobmanager::JobQueue::RunJob::runStep() 
jobmanager::WorkerThread::processJobs()
```

transport/Transport.h
```diff
@@ -37,6 +32,12 @@ class Transport
     virtual void connect() = 0;
     virtual jobmanager::JobQueue& getJobQueue() = 0;
     virtual void protectAndSend(memory::UniquePacket packet) = 0;
+
+    template <class Callable>
+    bool postOnQueue(Callable&& callableFunction)
+    {
+        return getJobQueue().post(getJobCounter(), std::forward<Callable>(callableFunction));
```
RicardoMDomingues, line 39: Utils function to post on JobQueue including the jobCounter

concurrency/SynchronizationContext.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include "concurrency/MpmcQueue.h"
+#include "utils/Function.h"
+
+namespace concurrency
+{
+
+class SynchronizationContext
```
RicardoMDomingues, line 9: This holds a reference to a queue but is no longer a queue. The goal is provide a way to add Functions to the queue without make available other methods like `pop`.

@olofkallander, I didn't called a queue because is no longer a queue, is more a wrapper to a queue where the push methods are available. I based the name on `SynchronizationContext` available on .net framework, please suggest a new name if you think this is not a good one

bridge/engine/EngineMixer.cpp
```diff
@@ -3379,7 +3633,7 @@ void EngineMixer::decommissionInboundContext(const uint32_t ssrc)
     {
         _ssrcInboundContexts.erase(ssrc);
         logger::info("Decommissioned inbound ssrc context %u", _loggableId.c_str(), ssrc);
-        it->second->sender->getJobQueue().addJob<RemoveInboundSsrcContextJob>(ssrc, *it->second->sender, *this);
+        it->second->sender->postOnQueue(utils::bind(&EngineMixer::internalRemoveInboundSsrc, this, ssrc));
```
RicardoMDomingues, line 3636: Job replaced with the generic method

bridge/engine/EngineMixer.cpp
```diff
@@ -4037,7 +4368,7 @@ void EngineMixer::removeBarbell(size_t idHash)
         decommissionInboundContext(audioStream.ssrc);
     }
 
-    barbell->transport.getJobQueue().addJob<RemoveBarbellJob>(*this, barbell->transport, barbell->idHash);
+    barbell->transport.postOnQueue(utils::bind(&EngineMixer::internalRemoveBarbell, this, barbell->idHash));
```
RicardoMDomingues, line 4371: Job replaced with the generic method

transport/TransportImpl.cpp
```diff
@@ -69,40 +70,6 @@ class PacketReceiveJob : public jobmanager::Job
...
-public:
-    UnregisterEndpointsJob(TransportImpl& transport) : CountedJob(transport.getJobCounter()), _transport(transport) {}
-
-    void run() override { _transport.internalUnregisterEndpoints(); }
-
-private:
-    TransportImpl& _transport;
-};
-
 class IceDisconnectJob : public jobmanager::Job
```
RicardoMDomingues, line 73: @olofkallander, I noticed this is no longer used. Did we remove the code by mistake or this not make sense any longer and I can remove?

bridge/engine/EngineMixer.cpp
```diff
@@ -66,117 +68,133 @@ memory::UniquePacket createGoodBye(uint32_t ssrc, memory::PacketPoolAllocator& a
...
-    transport::RtcTransport& _transport;
-    uint32_t _ssrc;
+    bridge::EngineMixer& _mixer;
+    bridge::SsrcOutboundContext& _outboundContext;
+    bridge::EngineMessageListener& _mixerManager;
+    size_t _endpointIdHash;
 };
 
-class SetRtxProbeSourceJob : public jobmanager::CountedJob
+class FinalizeNonSsrcRewriteOutboundContextJob : public jobmanager::CountedJob
```
RicardoMDomingues, line 109: Perhaps it would be better move this big jobs to a new cpp/h file

jobmanager/Job.h
olofkallander: can we put this comment above the CallableJob?

transport/TransportImpl.cpp
olofkallander: hmm, that is unfortunate. I wonder if there could be a utils::moveXX for this to avoid the stacking.

bridge/engine/EngineMixer.cpp
```diff
@@ -1850,14 +1931,33 @@ void EngineMixer::onRtcpPacketDecoded(transport::RtcTransport* sender,
...
+        if (ssrcOutboundContextItr->second.markedForDeletion)
+        {
+            assert(isForwardSsrc);
+            return nullptr;
+        }
+
         return &ssrcOutboundContextItr->second;
     }
 
+    if (isForwardSsrc)
```
olofkallander, line 1949: I am not sure about this flag. In a way it is making two methods out of one. We want a normal method to obtain outbound ssrc for rewrite capable streams right? I think it would be cleaner with another obtain method that does the stuff for non-rewrite streams. We want to get rid of them and there are mainly recording streams that use it.

bridge/engine/EngineMixer.cpp
```diff
@@ -2190,10 +2380,12 @@ void EngineMixer::forwardAudioRtpPacket(IncomingPacketInfo& packetInfo, uint64_t
                 ssrc = rewriteMapItr->second;
             }
 
+            const bool isForwardSsrc = !ssrcRewrite;
```
olofkallander, line 2383: I would prefer if there was another method to call to separate this. Also this variable is not really helping I think.

olofkallander:
> No, we must make sure it is really not needed. There is an onDisconnect method for a reason I think. We may be losing logging of this otherwise. 

Leave this for now, I will create a ticket to fix it.

olofkallander (approved):

PR #214, opened 2022-10-21
Limit forwarding to mid-quality when slids are being presented.

reddvl1980:
Add unit tests.

### DIFF
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -288,7 +288,8 @@ EngineMixer::EngineMixer(const std::string& id,
       _hasSentTimeout(false),
       _probingVideoStreams(false),
       _minUplinkEstimate(0),
-      _lastRecordingAckProcessed(utils::Time::getAbsoluteTime())
+      _lastRecordingAckProcessed(utils::Time::getAbsoluteTime()),
+      _slidesPresent(0)
 {
     assert(audioSsrcs.size() <= SsrcRewrite::ssrcArraySize);
     assert(videoSsrcs.size() <= SsrcRewrite::ssrcArraySize);
@@ -3832,8 +3833,8 @@ void EngineMixer::checkVideoBandwidth(const uint64_t timestamp)
     minUplinkEstimate =
         std::max(_config.slides.minBitrate.get(), std::min(minUplinkEstimate, getMinRemoteClientDownlinkBandwidth()));
 
-    uint32_t highQualityBitrate = 750;
-    if (presenterSimulcastLevel)
+    const bool slidesPresent = presenterSimulcastLevel;
+    if (slidesPresent)
     {
         const uint32_t slidesLimit = minUplinkEstimate * _config.slides.allocFactor;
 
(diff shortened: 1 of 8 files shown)


bridge/engine/EngineStreamDirector.h
```diff
@@ -475,7 +475,10 @@ class EngineStreamDirector
 
         const auto pinMapItr = _pinMap.find(toEndpointIdHash);
         const bool fromPinnedEndpoint = pinMapItr != _pinMap.end() && isSsrcFromParticipant(pinMapItr->second, ssrc);
-        const auto assignedQuality = (fromPinnedEndpoint ? viewer->pinQualityLevel : viewer->unpinQualityLevel);
+
+        // In case slides are used - limit video resolution to save CPU on recipients.
+        const auto assignedQuality =
+            (fromPinnedEndpoint && 0 == _slidesBitrateKbps) ? viewer->pinQualityLevel : viewer->unpinQualityLevel;
```
olofkallander, line 481: Interesting way of "detecting" slides presence. We do have the contentType in the adding of participant streams. But I guess this is more dynamically checking that slides are actually sent also, not just announced.

bridge/engine/EngineMixer.cpp
```diff
@@ -3851,8 +3852,28 @@ void EngineMixer::checkVideoBandwidth(const uint64_t timestamp)
...
+        auto& videoStream = *videoIt.second;
+        if (3 == videoStream.simulcastStream.numLevels)
+        {
+
+            logger::info("setting high-res bitrate for ssrc %u, at %u",
+                _loggableId.c_str(),
+                videoStream.simulcastStream.levels[2].ssrc,
+                highQualityBitrate);
+
+            presenterStream->transport.getJobQueue().addJob<SetMaxMediaBitrateJob>(videoStream.transport,
```
olofkallander, line 3870: post on videoStream instead

bridge/engine/EngineMixer.cpp
```diff
@@ -3851,8 +3852,28 @@ void EngineMixer::checkVideoBandwidth(const uint64_t timestamp)
...
+        highQualityBitrate = 8000;
         _engineStreamDirector->setSlidesSsrcAndBitrate(0, 0);
     }
+
+    for (auto videoIt : _engineVideoStreams)
+    {
+        auto& videoStream = *videoIt.second;
+        if (3 == videoStream.simulcastStream.numLevels)
+        {
+
```
olofkallander, line 3864: empty line.

bridge/engine/EngineMixer.cpp
```diff
@@ -3851,8 +3852,28 @@ void EngineMixer::checkVideoBandwidth(const uint64_t timestamp)
...
+            logger::info("setting high-res bitrate for ssrc %u, at %u",
+                _loggableId.c_str(),
+                videoStream.simulcastStream.levels[2].ssrc,
+                highQualityBitrate);
+
+            videoStream.transport.getJobQueue().addJob<SetMaxMediaBitrateJob>(videoStream.transport,
```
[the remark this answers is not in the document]
reddvl1980: Add a commit a second ago :)

RicardoMDomingues (approved):
General comment, maybe to other PR.

`Engine::checkVideoBandwidth` looks to me a method that only make sense to be called when we have screensharing. The we can save some CPU for conferences with video but without screensharing.

I think we could do:
- Send TMMBR to enable/disable high res when SLIDES is added/removed instead of sending in each 3 seconds. So we reduce the number of iteration again from 2N to N on this method.
- Store a flag on mixer to know if we have a SLIDE stream an only run this method if we do

I am thinking also maybe sending TMMBR could also be beneficial when we have calls 1-1. Instead of giving up of heigh res completely, perhaps we could adjust the bitrate as there is still only one participant to see it

PR #217, opened 2022-10-25
bbaudio

olofkallander:


RicardoMDomingues (approved):

bridge/engine/ActiveMediaList.cpp
```diff
@@ -130,14 +130,24 @@ bool ActiveMediaList::onAudioParticipantAdded(const size_t endpointIdHash, const
         _dominantSpeakerId = endpointIdHash;
     }
 
+    if (_audioSsrcRewriteMap.size() == _audioLastN)
```
RicardoMDomingues, line 133: This is important. We will have 2 more ssrcs available than the _audioLastN. We can't allow  to go beyond _audioSsrcRewriteMap.

If you don't limit this, it can grows until 5 and we can have until 10 audio changes on bb (5 endpoints removed and other new 5 added) but the map only has space to 8.

bridge/engine/EngineMixer.cpp
```diff
@@ -4169,7 +4169,7 @@ void EngineMixer::onBarbellUserMediaMap(size_t barbellIdHash, const char* messag
         }
     }
 
-    memory::Map<size_t, BarbellMapItem, 8> audioSsrcs;
+    memory::Map<size_t, BarbellMapItem, 10> audioSsrcs;
```
RicardoMDomingues, line 4172: As we have 5 audioSsrc I was thinking that were possible 10 changes. But I see now that is wrong. I will revert it

bridge/engine/EngineMixer.cpp
```diff
@@ -1019,6 +1019,7 @@ void EngineMixer::runDominantSpeakerCheck(const uint64_t engineIterationStartTim
     bool dominantSpeakerChanged = false;
     bool videoMapChanged = false;
     bool audioMapChanged = false;
```
olofkallander, line 1021: I think we can remove this flag altogether. We need to keep if videomapping changed as that is only reason to send UMM to clients.

reddvl1980: But what about UMM to barbell? W/o UMM after one endpoint has replaced another in the audio ssrc map, receiver will be confused and received audio will be attributed to the wrong endpoint, which will led to wrong decision on the active media list.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -310,23 +362,23 @@ bool ActiveMediaList::removeVideoParticipant(const size_t endpointIdHash)
...
-        participantLevels.maxRecentLevel -=
-            (participantLevels.maxRecentLevel - averageLevelLongWindow) * AudioParticipant::MAX_LEVEL_DECAY;
-        // Move old min level over time towards mean about 3dB per 3 seconds
-        participantLevels.noiseLevel = participantLevels.noiseLevel + AudioParticipant::NOISE_RAMPUP;
-        participantLevels.noiseLevel =
-            std::max(participantLevels.noiseLevel, static_cast<float>(AudioParticipant::MIN_NOISE));
+        audioParticipant.maxRecentLevel -=
+            (audioParticipant.maxRecentLevel - audioParticipant.noiseLevel) * AudioParticipant::MAX_LEVEL_DECAY;
+
+        if (audioParticipant.history.allNonZero() &&
```
olofkallander, line 375: problem that noise level could rise indefinitely if media muted.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -337,44 +389,24 @@ void ActiveMediaList::updateLevels(const uint64_t timestamp)
...
-        else if (levelEntry.level != 0 && participantLevels.nonZeroLevelsShortWindow == lengthShortWindow)
+        audioParticipant.maxRecentLevel =
+            std::max(audioParticipant.maxRecentLevel, static_cast<float>(levelEntry.level));
+        if (levelEntry.level == 0)
         {
-            participantLevels.noiseLevel = std::min(participantLevels.noiseLevel,
-                static_cast<float>(participantLevels.totalLevelShortWindow) / lengthShortWindow);
+            audioParticipant.maxRecentLevel = 0; // muted
         }
+        audioParticipant.history.update(levelEntry.level, timestamp);
```
olofkallander, line 401: moved level logic out

bridge/engine/ActiveMediaList.cpp
```diff
@@ -481,41 +542,39 @@ void ActiveMediaList::process(const uint64_t timestamp,
...
+        _nominatedSpeaker = nominatedSpeaker.participant;
     }
 
-    _lastRunTimestamp = timestamp;
-    if (dominantSpeaker.participant != _dominantSpeakerId &&
-        ((!_dominantSpeakerId || currentDominantSpeakerScore < 0.01) ||
-            (_consecutiveDominantSpeakerWins >= requiredConsecutiveWins &&
-                currentDominantSpeakerScore < 0.75 * dominantSpeaker.score &&
-                timestamp - _lastChangeTimestamp >= maxSwitchDominantSpeakerEvery)))
+    // switch at most every 2s
```
olofkallander, line 561: simplified condition. Dominant speaker gets grace period of 2s.
Bargein can happen either if dominant stops speaking or you try to bargein for 2s.

bridge/engine/EngineMixer.cpp
```diff
@@ -2747,14 +2748,14 @@ void EngineMixer::processIncomingBarbellFbRtcpPacket(EngineBarbell& barbell,
 
     auto& bbTransport = barbell.transport;
     auto* mediaSsrcOutboundContext =
-        obtainOutboundSsrcContext(barbell.idHash, barbell.ssrcOutboundContexts, mediaSsrc, barbell.audioRtpMap);
+        obtainOutboundSsrcContext(barbell.idHash, barbell.ssrcOutboundContexts, mediaSsrc, barbell.videoRtpMap);
```
reddvl1980, line 2751: How this even worked before?

olofkallander:
> How this even worked before? 
When we received rtcp fb the outboundcontext was already created as we had already sent rtp. For the RTX outbound context it may not have been created, but since we do not send RTP on that one we probably got away with incorrect rtp map.

PR #220, opened 2022-10-28
hiactlevel reset

olofkallander:
highest active levl must be proberly reset after BB resue of video stream. And also set at inactivity in case a stream level is lost over BB due to remote side dropping a level.

### DIFF
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -1149,15 +1149,15 @@ void EngineMixer::internalRemoveInboundSsrc(uint32_t ssrc)
         return;
     }
 
-    auto contextIt = _allSsrcInboundContexts.find(ssrc);
-    if (contextIt != _allSsrcInboundContexts.end())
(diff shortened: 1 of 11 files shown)


RicardoMDomingues (approved):

bridge/engine/EngineMixer.cpp
```diff
@@ -1165,9 +1165,17 @@ void EngineMixer::checkInboundPacketCounters(const uint64_t timestamp)
         if (!recentActivity && receiveCounters.packets > 5 && inboundContext.activeMedia &&
             inboundContext.rtpMap.format != RtpMap::Format::VP8RTX)
         {
+            inboundContext.activeMedia = false;
+            if (!_engineStreamDirector->streamActiveStateChanged(endpointIdHash, ssrc, false))
+            {
+                logger::warn("streamActiveStateChanged not set. Participant may be removed %zu ssrc %u",
```
reddvl1980, line 1171: director's `streamActiveChange` returns false from `setHighestActiveIndex` if there are no changes in state. So log is misleading.

PR #225, opened 2022-11-08
more &

olofkallander:
changed params to & where it made sense. Did not add to previous PR to keep change count down a bit.
Also fixed some mistakes with jobs on Transport that did not increase job counter.

I know there are a number of lmbda captures to fix.

### DIFF
--- a/bridge/Mixer.cpp
+++ b/bridge/Mixer.cpp
@@ -705,7 +705,7 @@ bool Mixer::removeDataStreamId(const std::string& id)
     return removeDataStream(endpointId);
 }
 
-void Mixer::engineAudioStreamRemoved(EngineAudioStream& engineStream)
+void Mixer::engineAudioStreamRemoved(const EngineAudioStream& engineStream)
 {
     std::lock_guard<std::mutex> locker(_configurationLock);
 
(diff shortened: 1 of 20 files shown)


bridge/Mixer.cpp
```diff
@@ -376,7 +376,7 @@ void Mixer::allocateAudioBuffer(uint32_t ssrc)
     auto audioBuffer = std::make_unique<EngineMixer::AudioBuffer>();
     auto* rawAudioBuffer = audioBuffer.get();
     _audioBuffers.emplace(ssrc, std::move(audioBuffer));
-    _engine.post(utils::bind(&EngineMixer::addAudioBuffer, _engineMixer.get(), ssrc, rawAudioBuffer));
+    _engine.post([&]() { _engineMixer->addAudioBuffer(ssrc, rawAudioBuffer); });
```
RicardoMDomingues, line 379: you are capturing everything by reference but `ssrc` and `rawAudioBuffer` are a stack values. When the lambda run they will be dangling references

RicardoMDomingues (commented):
I seem there is a problem of dangling references in many places. Perhaps you want to replace most of `[&]` by `[=]`

PR #228, opened 2022-11-11
RTC-13554: Fix SEGV and deadlock

RicardoMDomingues:


jobmanager/JobManager.h
```diff
@@ -121,13 +121,13 @@ class JobManager // TODO rename to MainJobQueue or MpmcJobQueue
     template <class Callable>
     bool post(Callable&& callable)
     {
-        return addJob<jobmanager::CallableJob<Callable>>(std::forward<Callable>(callable));
+        return addJob<jobmanager::CallableJob<std::decay_t<Callable>>>(std::forward<Callable>(callable));
```
RicardoMDomingues, line 124: We have to `decay` this or we most likely store references to stack objects

bridge/Bridge.cpp
```diff
@@ -153,7 +153,9 @@ void Bridge::initialize(std::shared_ptr<transport::EndpointFactory> endpointFact
     }
 
     startWorkerThreads();
-    backgroundWorker = std::make_unique<jobmanager::WorkerThread>(*_backgroundJobQueue, "MMWorker");
+    // Disabling yield because we don't want to yield jobs that holds mixer and MixerManager locks, otherwise it can create deadlocks
+    const bool yieldEnabled = false;
```
RicardoMDomingues, line 157: Yielding jobs that holds the configuration lock, can potential create deadlocks if another job that needs the lock is picked

reddvl1980 (approved):

PR #232, opened 2022-11-16
ep message fix

olofkallander:


api/DataChannelMessageParser.cpp
```diff
@@ -8,65 +7,69 @@ namespace DataChannelMessageParser
...
 } // namespace
 
-bool isPinnedEndpointsChanged(const nlohmann::json& messageJson)
+bool isPinnedEndpointsChanged(const utils::SimpleJson& messageJson)
 {
+    char messageType[200];
     if (isLegacyApi(messageJson))
     {
-        return legacyapi::DataChannelMessageParser::isPinnedEndpointsChanged(messageJson);
+        messageJson["colibriClass"].getString(messageType);
```
RicardoMDomingues, line 21: I think it would be nice if `getString()` could return an `utils::span<char>` so we could make the comparison with zero memory copies

RicardoMDomingues (approved):

PR #247, opened 2022-11-25
Olofproperdtlsfix

olofkallander:
several issues with reduced mem fottprint and udpendpoint
Mostly visible when running dtlsRace tests.

### DIFF
--- a/test/transport/IceIntegrationTest.cpp
+++ b/test/transport/IceIntegrationTest.cpp
@@ -274,7 +274,7 @@ TEST_F(IceIntegrationTest, dtlsRace)
     ASSERT_TRUE(_transportFactory2->isGood());
 
     std::vector<ClientPair*> clients;
-    for (int i = 0; i < 100; ++i)
+    for (int i = 0; i < 150; ++i)
     {
         clients.push_back(new ClientPair(*_transportFactory1,
             *_transportFactory2,
@@ -293,7 +293,7 @@ TEST_F(IceIntegrationTest, dtlsRace)
     }
 
     bool connected = false;
-    while (!connected && utils::Time::getAbsoluteTime() - startTime < 5 * utils::Time::sec)
+    while (!connected && utils::Time::getAbsoluteTime() - startTime < 10 * utils::Time::sec)
     {
         int64_t maxSleep = utils::Time::sec;
 
@@ -311,13 +311,20 @@ TEST_F(IceIntegrationTest, dtlsRace)
 
     int connectCount = 0;
     for (auto* clientPair : clients)
-
     {
         if (clientPair->isConnected())
         {
             ++connectCount;
         }
+        else
+        {
+            logger::warn("client not connected. %s -> %s",
+                "dtlsRace",
+                clientPair->_transport1->getLoggableId().c_str(),
+                clientPair->_transport2->getLoggableId().c_str());
+        }
     }
+    logger::flushLog();
     EXPECT_TRUE(connected);
     EXPECT_EQ(connectCount, clients.size());
     for (auto* clientPair : clients)
--- a/transport/BaseUdpEndpoint.cpp
+++ b/transport/BaseUdpEndpoint.cpp
@@ -1,60 +1,8 @@
 #include "transport/BaseUdpEndpoint.h"
+#include "utils/Function.h"
 
 namespace transport
 {
-namespace
-{
-using namespace transport;
-class SendJob : public jobmanager::Job
-{
-public:
-    explicit SendJob(BaseUdpEndpoint& endpoint) : _endpoint(endpoint) {}
-
-    void run() override { _endpoint.internalSend(); }
-
-private:
-    BaseUdpEndpoint& _endpoint;
-};
-class ReceiveJob : public jobmanager::Job
-{
-public:
-    ReceiveJob(BaseUdpEndpoint& endpoint, int fd) : _endpoint(endpoint), _fd(fd) {}
-
-    void run() override
-    {
-#ifdef __APPLE__
-        _endpoint.internalReceive(_fd, 1);
-#else
-        _endpoint.internalReceive(_fd, 400);
-#endif
-    }
-
-private:
-    BaseUdpEndpoint& _endpoint;
-    int _fd;
-};
-
-class StopPortJob : public jobmanager::Job
-{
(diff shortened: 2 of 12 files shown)


test/transport/IceIntegrationTest.cpp
```diff
@@ -293,7 +293,7 @@ TEST_F(IceIntegrationTest, dtlsRace)
     }
 
     bool connected = false;
-    while (!connected && utils::Time::getAbsoluteTime() - startTime < 5 * utils::Time::sec)
+    while (!connected && utils::Time::getAbsoluteTime() - startTime < 20 * utils::Time::sec)
```
olofkallander, line 296: will revert before final commit

transport/ice/IceSession.cpp
```diff
@@ -525,19 +525,6 @@ void IceSession::onResponseReceived(IceEndpoint* endpoint,
     {
         return;
     }
```
olofkallander, line 527: responses should not really contain USERNAME so we should not waste time on it.

transport/ice/IceSession.cpp
```diff
@@ -694,11 +687,6 @@ void IceSession::sendResponse(IceEndpoint* endpoint,
     {
         response.add(StunError(code, errorPhrase));
     }
```
olofkallander, line 689: We should not add USERNAME either to responses.

transport/BaseUdpEndpoint.cpp
```diff
@@ -66,8 +39,8 @@ BaseUdpEndpoint::BaseUdpEndpoint(const char* name,
     : _state(Endpoint::CLOSED),
       _name(name),
       _localPort(localPort),
-      _receiveJobs(jobManager, 16),
-      _sendJobs(jobManager, 16),
+      _receiveJobs(jobManager, maxSessionCount),
+      _sendJobs(jobManager, maxSessionCount),
```
olofkallander, line 43: Will rever to 16.

transport/BaseUdpEndpoint.cpp
```diff
@@ -66,8 +39,8 @@ BaseUdpEndpoint::BaseUdpEndpoint(const char* name,
     : _state(Endpoint::CLOSED),
       _name(name),
       _localPort(localPort),
-      _receiveJobs(jobManager, 16),
-      _sendJobs(jobManager, 16),
+      _receiveJobs(jobManager, maxSessionCount),
```
olofkallander, line 42: A small queue here is fine for receive jobs as they are like 2, but the queue is also used for unreg jobs and that can fill the queue causing receive to halt forever.

transport/TransportImpl.cpp
```diff
@@ -1065,8 +1065,10 @@ void TransportImpl::internalIceReceived(Endpoint& endpoint,
             return;
         }
 
-        if (_rtpIceSession->getState() == ice::IceSession::State::CONNECTING)
+        //logger::debug("ice received, ICE state %u", _loggableId.c_str(), _rtpIceSession->getState());
+        if (_rtpIceSession->getState() != ice::IceSession::State::CONNECTED)
```
olofkallander, line 1069: Not usre about this yet. Something goes wron un unreg so there is a ref count 1 on transportimpl.

transport/UdpEndpointImpl.cpp
```diff
@@ -75,7 +39,7 @@ void UdpEndpointImpl::sendStunTo(const transport::SocketAddress& target,
     uint64_t timestamp)
 {
     auto* msg = ice::StunMessage::fromPtr(data);
-    if (msg->header.isRequest() && !_dtlsListeners.contains(target) && !_iceResponseListeners.contains(transactionId))
+    if (msg->header.isRequest() && !_iceResponseListeners.contains(transactionId))
```
olofkallander, line 42: I turned to relying only on transaction ids for response routing. It is the right way.

transport/TransportImpl.cpp
```diff
@@ -1065,7 +1065,8 @@ void TransportImpl::internalIceReceived(Endpoint& endpoint,
             return;
         }
 
-        if (_rtpIceSession->getState() == ice::IceSession::State::CONNECTING)
+        const auto iceState = _rtpIceSession->getState();
+        if (iceState == ice::IceSession::State::CONNECTING || iceState == ice::IceSession::State::READY)
         {
             endpoint.registerListener(source, this);
```
olofkallander, line 1071: What happened was that a packet was inbound on the udp endpoint and fit the user name.  Ice session was already in idle and transport just unregistered. But the packet slipped in and caused registration and packet to go to idle ice session.

transport/UdpEndpointImpl.cpp
```diff
@@ -93,42 +64,43 @@ void UdpEndpointImpl::sendStunTo(const transport::SocketAddress& target,
...
-    logger::debug("unregister %p", _name.c_str(), listener);
+    LOG("unregister %p", _name.c_str(), listener);
     for (auto& item : _iceListeners)
     {
         if (item.second == listener)
         {
             _iceListeners.erase(item.first);
-            _receiveJobs.addJob<UnRegisterNotifyListenerJob>(*this, *listener);
+            listener->onUnregistered(*this);
+            break;
```
olofkallander, line 103: only one username is registered per transport

transport/UdpEndpointImpl.cpp
```diff
@@ -252,32 +239,61 @@ void UdpEndpointImpl::registerListener(const std::string& stunUserName, IEvents*
...
+        return;
     }
+
+    logger::warn("dtls listener swap on %s skipped. Already removed", _name.c_str(), srcAddress.toString().c_str());
 }
 
 void UdpEndpointImpl::focusListener(const SocketAddress& remotePort, IEvents* listener)
 {
-    return;
+    _receiveJobs.post([this, remotePort, listener]() {
```
olofkallander, line 283: I do not understand how skipping this code was helping. focusListener was not called in the failing test dtlsRace.

test/transport/IceIntegrationTest.cpp
olofkallander: It will exit as soon as they are connected. Giving them a bit more time since there are many connections.

RicardoMDomingues (approved):

PR #252, opened 2022-11-28
Fix hash equivalence of std::string and const char* - to fix pin endp…

reddvl1980:
…oint API

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -516,6 +516,7 @@ set(TEST_FILES
     test/memory/RingAllocatorTest.cpp
     test/utils/StringTokenizerTest.cpp
     test/utils/TrackerTest.cpp
+    test/utils/StdExtensionsTest.cpp
     test/memory/RingBufferTest.cpp
     test/memory/ListTest.cpp
     test/memory/ArrayTest.cpp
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -516,6 +516,7 @@ set(TEST_FILES
     test/memory/RingAllocatorTest.cpp
     test/utils/StringTokenizerTest.cpp
     test/utils/TrackerTest.cpp
+    test/utils/StdExtensionsTest.cpp
```
RicardoMDomingues, line 519: Don't forget to add this file to the commit

RicardoMDomingues (approved):

utils/StdExtensions.h
```diff
@@ -82,6 +82,12 @@ struct hash<char*>
     uint64_t operator()(const char* s, size_t len) const { return hashBuffer(s, len); }
 };
 
+template <>
+struct hash<const char*>
+{
+    uint64_t operator()(const char* s) const { return hash<char*>{}(const_cast<char*>(s)); }
```
RicardoMDomingues, line 88: you could avoid the `const_cast` if you change the current `utils::hash<char*> to utils::hash<const char*>` and make this one the non-const specialisation

RicardoMDomingues: 🤔  Or perhaps you don't need the `const_cast` here as `utils::hash<char *>` receives a `const char*`

PR #253, opened 2022-11-29
Finalize bundle transports once endpoint leaves

RicardoMDomingues:


reddvl1980:
@olofkallander , @reddvl1980  with my last two commits, auto-delete of transport started to work. But I see that from conference to conference we add ~3MB (to be exact 345060 - 342096) of "used memory".

PR #256, opened 2022-11-30
Unit tests

RicardoMDomingues:


transport/BaseUdpEndpoint.cpp
```diff
@@ -66,7 +66,7 @@ BaseUdpEndpoint::BaseUdpEndpoint(const char* name,
     : _state(Endpoint::CLOSED),
       _name(name),
       _localPort(localPort),
-      _receiveJobs(jobManager, 16),
```
RicardoMDomingues, line 69: We had a problem with this size. We need to figure out what is the reasonable values

olofkallander (approved):

PR #264, opened 2022-12-08
Improve load tests

reddvl1980:
- switch from 'fake' to 'opus' for generated audio. Should not load much, since we'll use not more than 5 producers;
- add analysis: frequencies are analyzed and matched against known producers. Amplitude analysis is off;
- add five jump starters for test: 'DISABLED_smbMegaHoot_1' to DISABLED_smbMegaHoot_5' ('DISABLED_smbMegaHoot' is the same as 'DISABLED_smbMegaHoot_1);
- add logging to see what was sent and received.

### DIFF
--- a/test/integration/emulator/SfuClient.h
+++ b/test/integration/emulator/SfuClient.h
@@ -116,7 +116,7 @@ public:
           _loggableId("client", id),
           _recordingActive(true),
           _ptime(ptime),
-          _audioType(Audio::None),
+          _sendAudioType(Audio::None),
           _expectedReceiveAudioType(Audio::None)
     {
     }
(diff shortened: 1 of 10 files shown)


test/integration/emulator/SfuClient.h
```diff
@@ -702,7 +705,7 @@ class SfuClient : public transport::DataReceiver
                         rtpHeader->ssrc.get(),
                         rtpMap,
                         sender,
-                        _audioType,
+                        _expectedReceiveAudioType == Audio::None ? _audioType : _expectedReceiveAudioType,
```
olofkallander, line 708: ??

reddvl1980: For backward compatibilty - if setter is not called, we'll expect _audioType. Otherwise - new _expectedREceiveAudioType

olofkallander (approved):

olofkallander:
> For backward compatibilty - if setter is not called, we'll expect _audioType. Otherwise - new _expectedREceiveAudioType
 
ok, it is just if you read it as freshmen in the code, I wonder if it makes sense to them. 
Either a comment or perhaps the audioType should rename as sendAudioType or something.

PR #271, opened 2022-12-13
RTC-13657 Fix bug in EngineStreamDirector

reddvl1980:
Fix bug in EngineStreamDirector where lowest config with the same cost was chosen:
e.g. 2 participants video call, bitrate below 2000 kbps used to result in low/low config instead of mid/low
since both has starting 'cost' of 500

### DIFF
--- a/bridge/engine/EngineStreamDirector.h
+++ b/bridge/engine/EngineStreamDirector.h
@@ -996,7 +996,8 @@ private:
             assert(configCost >= config.MinBitrateMargin + _slidesBitrateKbps);
             assert(configCost <= config.MaxBitrateMargin + _slidesBitrateKbps);
 
-            if (configCost >= bestConfigCost && configCost <= estimatedUplinkBandwidth)
+            const auto configIsBetter = bestConfigCost == 0 || configCost > bestConfigCost;
+            if (configIsBetter && configCost <= estimatedUplinkBandwidth)
             {
                 bestConfigCost = configCost;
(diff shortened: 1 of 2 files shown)


bridge/engine/EngineStreamDirector.h
```diff
@@ -996,7 +996,8 @@ class EngineStreamDirector
             assert(configCost >= config.MinBitrateMargin + _slidesBitrateKbps);
             assert(configCost <= config.MaxBitrateMargin + _slidesBitrateKbps);
 
-            if (configCost >= bestConfigCost && configCost <= estimatedUplinkBandwidth)
+            const auto configIsBetter = bestConfigCost == 0 || configCost > bestConfigCost;
```
olofkallander, line 999: const auto configIsBetter = (bestConfigCost == 0 || configCost > bestConfigCost);

more readable

bridge/engine/EngineStreamDirector.h
```diff
@@ -996,7 +996,8 @@ class EngineStreamDirector
...
+            if (configIsBetter && configCost <= estimatedUplinkBandwidth)
```
olofkallander, line 1000: Maybe another PR but it is a bit confusing with the calculation above that OH bitrate is actual full bitrate of video. I thought first, why do we not multiple the number of streams with some video bit rate, and that is actually what is being done called overhead.

olofkallander (approved):

PR #273, opened 2022-12-15
Reopen log file if deleted (e.g. during log rotation)

reddvl1980:


logger/LoggerThread.cpp
```diff
@@ -57,6 +64,18 @@ void logStack(const LogItem& item, const char* localTime, bool logStdOut, FILE*
 }
 } // namespace
 
+void LoggerThread::ensureLogFileExists()
+{
+    if (_logFileName.size())
+    {
+        std::ifstream f(_logFileName.c_str());
```
RicardoMDomingues, line 71: I think this is an expensive why to check if a file exist as internally it will open a new file descriptor and initialise an `ifstream` and potential allocate some buffers on these operations. 
I believe a cheaper way to do this is by using `fstat` with file descriptor we already have and you can check how many hard links are for that inode (`st_nlink`). If it is zero, means the file was deleted.

reddvl1980: Switched to C++17 std::filesystem::exist

docker/el7/Dockerfile
```diff
@@ -40,8 +68,19 @@ RUN cd /tmp && wget https://archive.mozilla.org/pub/opus/opus-1.3.1.tar.gz \
     && make -j5 \
     && make install
 
+# Restore old GCC's ld to build glibc
+RUN ln -sf /usr/bin/ld /usr/local/bin/ld
+
 RUN cd /tmp && wget http://ftp.gnu.org/gnu/glibc/glibc-2.18.tar.gz && tar zxvf glibc-2.18.tar.gz \
     && cd /tmp/glibc-2.18 && mkdir -p build && cd build \
-    && CC= CXX= LD_LIBRARY_PATH= ../configure --prefix=/opt/glibc-2.18 \
+    && CC=gcc CXX=gcc++ LD_LIBRARY_PATH= ../configure --prefix=/opt/glibc-2.18 \
```
RicardoMDomingues, line 76: I think `gcc++` does not exist (probably is not a problem because glibc is not a `C++` project. The GNU C++ compiler is `g++`

```suggestion
    && CC=gcc CXX=g++ LD_LIBRARY_PATH= ../configure --prefix=/opt/glibc-2.18 \
```

docker/el7/Dockerfile
```diff
@@ -2,17 +2,45 @@ FROM centos:7
...
+# Build libcxx to /tmp/libcxx-8.0.1.src/build/lib/libc++.so.1.0 with GLIBC-2.17
+RUN cd /tmp \
+    && wget -O libcxx-8.0.1.src.tar.xz https://github.com/llvm/llvm-project/releases/download/llvmorg-8.0.1/libcxx-8.0.1.src.tar.xz \
+    && tar xvf libcxx-8.0.1.src.tar.xz \
+    && cd libcxx-8.0.1.src \ 
+    && source scl_source enable devtoolset-9 \
+    && mkdir build \
+    && cd build \
+    && cmake ../ -DCMAKE_BUILD_TYPE=release -G "Unix Makefiles" . \
+    && make -j$(getconf _NPROCESSORS_ONLN)
```
RicardoMDomingues, line 30: This will let the libraries in `tmp` directory which I think is not a good practice. Can you copy/install them in a permanent directory? If someone wants to use this image and robots the container I think it is not ensured the files will remains there

docker/el7/buildscript.sh
```diff
@@ -42,12 +41,13 @@ if [ ! -d libs ]; then
   mkdir libs
 fi
 
-cp /usr/local/lib/libc++.so.1 libs
-cp /usr/local/lib/libc++abi.so.1 libs
+cp /tmp/libcxx-8.0.1.src/build/lib/libc++.so.1.0 libs/libc++.so.1
+cp /tmp/libcxxabi-8.0.1.src/build/lib/libc++abi.so.1.0 libs/libc++abi.so.1
```
RicardoMDomingues, line 45: Please don't use the tmp directory

RicardoMDomingues (approved):

PR #276, opened 2023-01-09
formatTime

olofkallander:


logger/LoggerThread.h
```diff
@@ -31,7 +31,7 @@ class LoggerThread
 
 private:
     void run();
-    void formatTime(const LogItem& item, char* output);
+    static void formatTime(const std::chrono::system_clock::time_point timestamp, char* output);
```
RicardoMDomingues, line 34: private static methods are good candidates to be placed in anonymous namespaces

RicardoMDomingues (approved):

logger/LoggerThread.h
olofkallander: yes, it was supposed to be public so it can be used outside. Fixed that.

PR #277, opened 2023-01-10
reopen log on signal from logrotate

olofkallander:


olofkallander:
log rotation can be configured in different ways, and some of them allows for loss of logs.
1. copytruncate mode: logrotate copies the content of the file to a backup file and then truncates the original file. SMB works fine with this mode as it just continues to write to the existing file handle. However, logs written after copy, before truncate, will be lost. Small time frame, but still.

2. create mode: logrotate renames the original file, while smb can continue writing to it. The logrotate script should then send a USR1 signal to smb which causes it to reopen the smb.log, which will be created first by logrotate. In this mode we will lose nothing as the renamed file picks up the logs while rotation happens.

3. create mode with compress: Same as 2 but now logrotate renames, then zips the content of the renamed file, then it creates a new file and send USR1 signal. Problem is that any logs going to the "renamed file" after zip has been created, will be lost into a "deleted file". After signal has been picked up smb will switch to new smb.log but some logs may be lost at the end of compression.

I have tested all cases above.

PR #279, opened 2023-01-16
Opuslevel

olofkallander:
Calculate and set audio level on forwarded packets, when needed.

### DIFF
--- a/bridge/engine/ActiveMediaList.cpp
+++ b/bridge/engine/ActiveMediaList.cpp
@@ -979,7 +979,7 @@ void ActiveMediaList::removeFromRewriteMap(size_t endpointIdHash)
 
 bool ActiveMediaList::isAudioLevelNeeded() const
 {
-    return _audioSsrcs.size() > _audioLastN;
+    return _audioParticipants.size() > _audioLastN;
 }
 
 #if DEBUG
--- a/bridge/engine/ActiveMediaList.cpp
+++ b/bridge/engine/ActiveMediaList.cpp
@@ -417,11 +417,7 @@ void ActiveMediaList::updateLevels(const uint64_t timestamp)
         audioParticipant.ptt = levelEntry.ptt;
         audioParticipant.onNewLevel(levelEntry.level, timestamp);
 
-        if (audioParticipant.ptt)
-        {
-            audioParticipant.noiseLevel = 37;
-        }
-        else if (audioParticipant.history.allNonZero())
+        if (audioParticipant.history.allNonZero())
         {
             audioParticipant.noiseLevel = std::min(audioParticipant.noiseLevel, audioParticipant.history.average());
         }
@@ -521,6 +517,8 @@ void ActiveMediaList::process(const uint64_t timestamp,
     size_t speakerCount = rankSpeakers();
     if (speakerCount == 0)
     {
+        TActiveTalkersSnapshot activeTalkersSnapshot;
+        _activeTalkerSnapshot.write(activeTalkersSnapshot);
         return;
     }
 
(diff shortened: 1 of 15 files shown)


bridge/engine/ActiveMediaList.cpp
```diff
@@ -977,6 +977,11 @@ void ActiveMediaList::removeFromRewriteMap(size_t endpointIdHash)
     ++_ssrcMapRevision;
 }
 
+bool ActiveMediaList::isAudioLevelNeeded() const
+{
+    return _audioParticipants.size() > _audioLastN;
```
olofkallander, line 982: I lot of reading of the atomic in LockFreeList but we avoid decoding opus and calculating audio level in 1-1 calls and small conferences < ~3 people.
Except for BB calls as we do not know how many are on the remote side and assume auido level is needed for correct barge-in decisions.

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -150,32 +305,19 @@ void AudioForwarderReceiveJob::run()
...
-
-            _engineMixer.mapSsrc2UserId(_ssrcContext.ssrc, c9UserId);
-        }
-
-        _activeMediaList.onNewAudioLevel(_packet->endpointIdHash,
-            audioLevel.valueOr(120),
-            isPtt.isSet() && isPtt.get());
+    if (audioLevel.isSet())
+    {
+        _activeMediaList.onNewAudioLevel(_packet->endpointIdHash, audioLevel.get(), isPtt.isSet() && isPtt.get());
```
olofkallander, line 320: If we have audio level ext we can optimize on near silent packets.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -538,16 +536,13 @@ void ActiveMediaList::process(const uint64_t timestamp,
         outAudioMapChanged |= updateActiveAudioList(top.participant);
         auto const& curParticipant = _audioParticipants.find(top.participant);
 
-        if (top.score - top.noiseLevel > _activeTalkerSilenceThresholdDb)
+        if (top.score > _activeTalkerSilenceThresholdDb && activeTalkersSnapshot.count < activeTalkersSnapshot.maxSize)
```
olofkallander, line 539: score is recent - noise , so I am not sure what this active talker selection was menat to be. One neede to be 18dB above 2x noise level to get into list.

bridge/engine/ActiveMediaList.cpp
```diff
@@ -521,6 +517,8 @@ void ActiveMediaList::process(const uint64_t timestamp,
     size_t speakerCount = rankSpeakers();
     if (speakerCount == 0)
     {
+        TActiveTalkersSnapshot activeTalkersSnapshot;
```
olofkallander, line 520: If we do not set here we will have a list with at least one active talker though all are muted.

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -188,34 +330,43 @@ void AudioForwarderReceiveJob::run()
...
-        logger::error("Failed to unprotect srtp %u, mixer %s",
-            "AudioForwarderReceiveJob",
-            _ssrcContext.ssrc,
-            _engineMixer.getLoggableId().c_str());
-        return;
+        calculatedAudioLevel = decodeOpus(*_packet, !audioLevel.isSet());
+    }
+    else if (_ssrcContext.rtpMap.format == bridge::RtpMap::Format::OPUS && _needAudioLevel && !audioLevel.isSet())
+    {
+        calculatedAudioLevel = computeOpusAudioLevel(*_packet);
```
olofkallander, line 345: If we have to make this feature configurable i guess we need to do something here to emulate the audio level on ptt.

RicardoMDomingues (approved):

PR #284, opened 2023-01-24
mpsc queue

olofkallander:
reduce mem footprint by 100MB
log with mpsc queue

### DIFF
--- a/concurrency/MpscMemoryQueue.h
+++ b/concurrency/MpscMemoryQueue.h
@@ -1,0 +1,155 @@
+#pragma once
+#include "utils/Allocator.h"
+#include <algorithm>
+#include <atomic>
+#include <cassert>
+#include <memory>
+
+namespace concurrency
+{
+/**
+ * Multiple producer single consumer queue of arbitrary large memory blocks.
+ * Useful as wait free queue for variable size items in mpsc setting.
+ */
+class MpscMemoryQueue
+{
+    enum CellState : uint32_t
+    {
+        emptySlot = 0,
+        allocated,
+        committed,
+        padding
+    };
+
+    struct Guard
+    {
+        Guard() {}
+        explicit Guard(bool allocated)
+        {
+            if (allocated)
+            {
+                std::memcpy(pattern, "ABXYGAHAERLKBOSP", 16);
+            }
+        }
+
+        bool operator==(const Guard& o) const { return 0 == std::memcmp(pattern, o.pattern, sizeof(pattern)); }
+
+        char pattern[16] = {0};
+    };
+
+    struct Entry
+    {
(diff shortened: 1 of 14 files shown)


concurrency/MpscMemoryQueue.h
```diff
@@ -0,0 +1,155 @@
...
+
+        void checkGuards() const;
+        Guard& tailGuard() { return *reinterpret_cast<Guard*>(data + size); }
+        const Guard& tailGuard() const { return *reinterpret_cast<const Guard*>(data + size); }
+#ifdef DEBUG
+        Guard frontGuard;
+#endif
+        std::atomic<CellState> state;
+        uint32_t size;
+        uint8_t data[];
```
RicardoMDomingues, line 67: This only works for structures with alignment <= 32 bits

concurrency/MpscMemoryQueue.h
```diff
@@ -0,0 +1,155 @@
...
+{
+public:
+    ScopedAllocCommit(MpscMemoryQueue& heap, uint32_t size) : _heap(heap), _committed(false)
+    {
+        _ptr = heap.allocate(size);
+    }
+    ~ScopedAllocCommit() { commit(); }
+
+    explicit operator bool() noexcept { return _ptr != nullptr; }
+    void* operator->() { return _ptr; }
```
RicardoMDomingues, line 133: add this operator to return a `void*` is pointless as we can't access to any member of a `void*`

concurrency/MpscMemoryQueue.h
```diff
@@ -0,0 +1,155 @@
...
+#include <cassert>
+#include <memory>
+
+namespace concurrency
+{
+/**
+ * Multiple producer single consumer queue of arbitrary large memory blocks.
+ * Useful as wait free queue for variable size items in mpsc setting.
+ */
+class MpscMemoryQueue
```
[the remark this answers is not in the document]
olofkallander: I thought of it as a more generic queue of varying size memory blocks where types may actually vary. But if the type does vary, the user would have to implement some sort of dynamic type identifier to know how to cast the generic objects in the queue. Perhaps it is better to  use a base type with size and type id as that is most likely the use case anyway. And return that base type....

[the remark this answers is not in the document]
olofkallander: Intention was to have a generic memory block where type size varies and could even be different. But that is achievable I guess even if this is a template. It is just that the whole purpose is that the size of the type may vary.

concurrency/MpscMemoryQueue.h
olofkallander: will fix

RicardoMDomingues (approved):

PR #285, opened 2023-01-25
alignment

olofkallander:
fix alignment checks

### DIFF
--- a/concurrency/MpscQueue.h
+++ b/concurrency/MpscQueue.h
@@ -34,7 +34,6 @@ class MpscQueueBase
 
         Entry() : state(State::emptySlot), size(0) {}
         size_t entrySize() const { return entryOverHead() + size; }
-        static constexpr uint32_t headSize() { return sizeof(Entry); }
         static constexpr uint32_t entryOverHead()
         {
 #ifdef DEBUG
@@ -59,7 +58,7 @@ class MpscQueueBase
         uint32_t size;
         uint8_t data[];
     };
-    static_assert(sizeof(Entry) % alignof(uint64_t) == 0, "MpscQueueBase::Guard must allow 64bit alignment of data");
+    static_assert(sizeof(Entry) % alignof(uint64_t) == 0, "Entry must allow 64bit alignment of data");
 
     struct CursorState
     {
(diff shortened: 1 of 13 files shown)


concurrency/MpscQueue.h
```diff
@@ -55,9 +56,9 @@ class MpscQueueBase
 #endif
         std::atomic<State> state;
         uint32_t size;
-        uint32_t _align32;
         uint8_t data[];
```
RicardoMDomingues, line 59: Alignment should be explicit. Otherwise it's fragile.
You know that it's aligned because we had 2x32bit field in front and you are ensuring that we will store it in a 64bit aligned memory. But If you create an `Entry` on stack memory for example, it will requires only 32bit alignment, also is very sensitive to new fields that we can add to `Entry` in the future.

Making explicitly aligned to 64bits. Will automatically make the `Entry` to be aligned to 64bit and it's ensured by compiler that will be correct if you add new fields in the future

```suggestion
        alignas(64) uint8_t data[];
```

concurrency/MpscQueue.h
```diff
@@ -55,9 +56,9 @@ class MpscQueueBase
...
     };
+    static_assert(sizeof(Entry) % alignof(uint64_t) == 0, "Entry must allow 64bit alignment of data");
```
RicardoMDomingues, line 61: If you follow my previous suggestion you can add this static_assert:
`static_assert(alignof(Entry) == alignof(uint64_t))`;

RicardoMDomingues (approved):

concurrency/MpscQueue.h
RicardoMDomingues: Actually I am a bit surprised how how that works as the struct requires 32bit alignment. I was expecting the size to be 12 in release 🤔

RicardoMDomingues: I make some experiments, interesting that `data[]`is considered an incomplete type and seems the sizeof Entry is 8. But the requirement alignment is still 32bits. I wondering if `uint8_t data[]` is actually the best data type to use in this case. Perhaps a `void*` could be better and we know it would be 64bit aligned

PR #286, opened 2023-01-31
MpmcQ uplift

olofkallander:
- arbitrary queue size allowed, which helps reducing memory footprint
- 25% performance improvement due to 
  - Fewer atomic operations.
  - read independant from write: less contention 
  - Better layout of members
  - dispersed access pattern of the queue avoids cache line 
     contention on elements.
- allow items without default constructor as items are created in 
   place when needed.

### DIFF
--- a/concurrency/MpmcQueue.h
+++ b/concurrency/MpmcQueue.h
@@ -5,40 +5,86 @@
 #include <cassert>
 namespace concurrency
 {
-// It is wise to make your queue entries at least 56B large
-// to facilitate cache line separation
+
+// High performing thread safe Multiple Producer, Multiple Consumer, Wait-free queue but...
+// If a reader is pre-empted when reading the element and a writer catches up, the writer will fail to push as the write
+// slot is still being read. This happens typically when the queue is rather full and there are competing threads in the
+// system.
 template <typename T>
 class MpmcQueue
 {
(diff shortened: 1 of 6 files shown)


concurrency/MpmcQueue.h
```diff
@@ -5,112 +5,164 @@
...
-                // A reader that wrapped will not pass readable test because the pos == writepos
-                auto& entry = _elements[pos % _maxElements];
-                target = std::move(entry.value);
-                entry.state.store(CellState::emptySlot, std::memory_order_release);
+                // A reader that wrapped will not pass readable test because the readcursor version does not match cell
+                // state version
+                auto& entry = _elements[indexTransform(pos)];
+                target = std::move(entry.value());
+                entry.value().~T();
+                entry.state.store(CellState{pos.version + 1, CellState::emptySlot}, std::memory_order_release);
```
olofkallander, line 134: This is the Achilles heel of MpmcQueue. There is a lock-read-release phase here that can stop writers from continuing.
If the element would have been removed from the queue in one atomic operation, the reader would either succeed in acquiring the object or fail to retry. Instead of the item containing value, the queue should have contained a versioned pointer.

olofkallander:
Will reconsider solution to make it truly wait free.

PR #287, opened 2023-02-01
mpmc mem footprint

olofkallander:
arbitrary queue size allowed, which helps reducing memory footprint
25% performance improvement due to
Fewer atomic operations.
read independant from write: less contention
Better layout of members
dispersed access pattern of the queue avoids cache line
contention on elements.
allow items without default constructor as items are created in
place when needed.

### DIFF
--- a/concurrency/MpmcQueue.h
+++ b/concurrency/MpmcQueue.h
@@ -5,40 +5,86 @@
 #include <cassert>
 namespace concurrency
 {
-// It is wise to make your queue entries at least 56B large
-// to facilitate cache line separation
+
+// High performing thread safe Multiple Producer, Multiple Consumer, Wait-free queue but...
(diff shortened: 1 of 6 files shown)


concurrency/MpmcQueue.h
```diff
@@ -5,112 +5,164 @@
...
+
     struct Entry
     {
-        Entry() : state(emptySlot) {}
+        Entry() {}
 
+        T& value() { return reinterpret_cast<T&>(data); }
         std::atomic<CellState> state;
-        T value;
+        uint8_t data[sizeof(T)];
```
RicardoMDomingues, line 32: 32bits in front and no alignment restriction  on `data`. Seems safe for `T` with alignment requirement <= 32bits

RicardoMDomingues (approved):

PR #289, opened 2023-02-13
Add optional H264 support

marcusspangenberg:
- Optional H264 support. Not negotiated since we are forwarding packets. SMB offers one or the other by selecting video codec in the config file for the application lifetime. Default is VP8
- Packetization mode 0 and 1 are supported
- Unit tests added for header and rewriter

Tested manually with and without simulcast in example/simpleclient as well as with Eyevinn WHIP broadcast components.

@reddvl1980 Since you are the H264 expert here, please let me know if I missed anything regarding keyframe detection for mode 0 and 1.


codec/H264Header.h
```diff
@@ -0,0 +1,64 @@
...
+constexpr uint8_t getFuNalStartBit(const uint8_t fuHeader)
+{
+    return fuHeader & 0x80;
+}
+
+inline bool isKeyFrameStapA(const uint8_t* payload, const size_t payloadSize)
+{
+    size_t offset = 1;
+
+    while (payloadSize - offset >= 3)
```
olofkallander, line 28: There is no check that the payloadSize is > 0 in callstack to this method. I would prefer if this method return false if ever (payloadSize < offset);

olofkallander:
LGTM. Maybe we could consider later adding a VideoSourc in the Integration test parts to produce fake H264 packets with valid headers.

codec/H264Header.h
marcusspangenberg: Thanks, I fixed the incorrect length check + added a test for malformed stap-a packet. I also inlined the function to make it clear it is not meant to be called separately.

olofkallander:
CI failed in tcheck (likely due to really slow thread scheduling)
It failed in release due to a tcp connectivity problem, which we see occasionally.
It also failed in addr sanitizerrr and that one looked a bit more suspicious. I am re-running now, but it may be something we need to investigate.

marcusspangenberg:
> CI failed in tcheck (likely due to really slow thread scheduling) It failed in release due to a tcp connectivity problem, which we see occasionally. It also failed in addr sanitizerrr and that one looked a bit more suspicious. I am re-running now, but it may be something we need to investigate.

Seems to pass now. Ok to merge, or was the asan error in the new code? In that case I can try to run that locally.

marcusspangenberg:
Ok to merge?

marcusspangenberg:
@olofkallander did you see anything that needs to be addressed? If everything looks ok, could you approve the PR? Thanks.

RicardoMDomingues (approved):
LGTM

PR #290, opened 2023-02-13
Accept ICE candidates beyond CONNECT

olofkallander:
intended changes:
- ICE probes, resulting in new candidates / sources, are accepted throughout the session.
- Those candidates are also probed by SMB
- Candidates that lose connectivity will be discarded, which saves slots in udpEP
- IceSession actually continues to probe all candidates in list, which makes it possible to test client side of ICE.
- Ice tests with emulated time now use TimeTurner so we get the logs timestamped accordingly.
- I relaxed the power of 2 requirement on MpmcHashmap to allow more precise sizing.
part of MEDIA-1718

Not done and left for another ticket: 
- select another remote candidate if the one used stops working and is discarded.
- select candidate based on RTT, prio, nomination now that we have probing. It is not trivial.

### DIFF
--- a/test/transport/IceTest.cpp
+++ b/test/transport/IceTest.cpp
@@ -545,9 +545,9 @@ TEST_F(IceTest, gather)
         }
     }
 
-    utils::Time::initialize();
+    std::thread t([&infra]() { infra.stop(); });
     timeSource.shutdown();
-    infra.stop();
+    t.join();
 }
 
 typedef std::vector<std::unique_ptr<ice::IceSession>> IceSessions;
--- a/test/transport/IceTest.cpp
+++ b/test/transport/IceTest.cpp
@@ -761,7 +761,7 @@ void logStatus(IceSessions& sessions)
     {
         logger::info("session %d local candidates", "", si);
         log(session->getLocalCandidates());
-        logger::info("session %d remote candidates", "", si++);
+        logger::info("session %d remote candidates", "", si);
         log(session->getRemoteCandidates());
 
         if (session->getState() == ice::IceSession::State::CONNECTED)
@@ -771,6 +771,7 @@ void logStatus(IceSessions& sessions)
             log(selectedPair1.first, "local");
             log(selectedPair1.second, "remote");
         }
+        ++si;
     }
 }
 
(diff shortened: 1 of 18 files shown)


transport/UdpEndpointImpl.cpp
```diff
@@ -29,7 +29,7 @@ UdpEndpointImpl::UdpEndpointImpl(jobmanager::JobManager& jobManager,
     bool isShared)
     : BaseUdpEndpoint("UdpEndpoint", jobManager, maxSessionCount, allocator, localPort, epoll, isShared),
       _iceListeners(maxSessionCount * 2),
-      _dtlsListeners(maxSessionCount * 2),
+      _dtlsListeners(maxSessionCount * 5),
```
olofkallander, line 32: Maybe discuss this. I reasoned that SMB sees only successful candidates here. Client may have many but usually only a couple of them will actually reach SMB. So even if a few clients do have more, most will connect with one successful candidate. Makes sense?

transport/TransportImpl.cpp
```diff
@@ -2328,4 +2333,27 @@ void TransportImpl::setTag(const char* tag)
     utils::strncpy(_tag, tag, sizeof(_tag));
 }
 
+void TransportImpl::onIceDiscardCandidate(ice::IceSession* session,
+    ice::IceEndpoint* endpoint,
+    const transport::SocketAddress& sourcePort)
+{
+    for (auto& udpEndpoint : _rtpEndpoints)
```
RicardoMDomingues, line 2340: I am wondering if this lookup is really necessary. What bad could happen if you call the `dpEndpoint->unregisterListener(sourcePort, this);` without check if it exists in `_rtpEndpoints` or ìn `_rtcpEndpoints`

transport/UdpEndpointImpl.cpp
RicardoMDomingues: Yes it make sense, but if will have successful candidates only,  isn't the `_dtlsListeners` most likely smaller then `_iceListeners`? Perhaps  I misunderstood it

olofkallander: The iceListener is listening for user id, pwd which is one per user.
The dtls Listener can be multiple per user as it is one per candidate.

test/transport/IceTest.cpp
```diff
@@ -807,18 +829,46 @@ bool establishIce(fakenet::NetworkNode& internet, IceSessions& sessions, uint64_
...
-            timeSource += timeout + 2;
+            timeSource.advance(timeout + 2);
         }
     }
 
     return !running;
 }
 
-TEST(IceTest, iceprobes)
+bool runIce(fakenet::NetworkNode& internet, IceSessions& sessions, utils::TimeSource& timeSource, uint64_t runTime)
```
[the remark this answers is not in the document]
olofkallander: probably could be changed will do next time.

transport/TransportImpl.cpp
olofkallander: I would still have to go through all endpoints and call unregister, then the endpoint would ignore it because it cannot find the sourceport, perhaps.

PR #297, opened 2023-03-28
MEDIA-1770 opus stats

olofkallander:
opus decode stats fixed

### DIFF
--- a/bridge/engine/AudioForwarderReceiveJob.cpp
+++ b/bridge/engine/AudioForwarderReceiveJob.cpp
@@ -349,17 +349,21 @@ void AudioForwarderReceiveJob::run()
     }
 
     int calculatedAudioLevel = -1;
-    if (_hasMixedAudioStreams && _ssrcContext.rtpMap.format == bridge::RtpMap::Format::OPUS)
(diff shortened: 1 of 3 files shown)


utils/Trackers.h
```diff
@@ -43,6 +43,8 @@ class AvgRateTracker
 
     void update(double value, uint64_t timestamp);
     double get() const;
+    double get(uint64_t timestamp, uint64_t maxSilence) const;
```
reddvl1980, line 46: I understand what the "maxSilence" is designed for, but consider changing the name to be less 'audio' and more generic. E.g. maxTimeWithOutSamples (or something shorter).

bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -359,6 +357,10 @@ void AudioForwarderReceiveJob::run()
     {
         calculatedAudioLevel = computeOpusAudioLevel(*_packet);
     }
+    else if (_ssrcContext.opusPacketRate)
+    {
+        _ssrcContext.opusPacketRate->set(0, 0);
```
reddvl1980, line 362: Do we want to log this case (may be once)?

reddvl1980 (approved):
LGTM (but read the comments).

PR #298, opened 2023-03-30
Feature/MEDIA-1774

olofkallander:
Fix to have the get the short list of conference contents, confid and endpoint ids.
I also detected a flaky test MpmcPublish that could fail when run in tcheck. This was due to writer not finding an empty slot when tcheck paused all threads in reading and writing. This is a good test so I improved the implementation of MpmcPublish read/write will always succeed when the threads are at most as many as specified.

### DIFF
--- a/concurrency/MpmcPublish.h
+++ b/concurrency/MpmcPublish.h
@@ -1,5 +1,7 @@
 #pragma once
 #include <atomic>
+#include <cassert>
+#include <iterator>
 
 namespace concurrency
 {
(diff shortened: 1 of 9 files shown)


concurrency/MpmcPublish.h
```diff
@@ -14,82 +16,77 @@ class MpmcPublish
...
     // return false if empty
     bool read(T& target) const
     {
-        if (empty())
-        {
-            return false;
-        }
-
-        for (;;)
+        for (auto* cell = _readPointer.load(std::memory_order_consume); cell;
```
olofkallander, line 35: Simplified a bit. Most likely outcome is that we get a readable item with ref count > 0.
One less load of the _readPointer as we check for empty from same read.

concurrency/MpmcPublish.h
```diff
@@ -14,82 +16,77 @@ class MpmcPublish
...
 
     // iterate until we find an idle slot to write into.
     // most reading threads will be locking the previous write pos
     bool write(const T& obj)
     {
-        uint32_t readCursor = _readCursor.load(std::memory_order::memory_order_relaxed);
-        const uint32_t readIndex = indexOf(readCursor);
-        const uint32_t version = versionOf(readCursor);
-        for (int i = 0; i < MAX_THREADS - 1; ++i)
+        auto readPointer = _readPointer.load(std::memory_order_consume);
```
olofkallander, line 56: Simpler and more efficient write. In most likely case there are  load + fetchadd + cmpexch.
Before it had to: load, cmpexch, fetchsub, cmpexch, fetchsub

RicardoMDomingues (approved):

PR #301, opened 2023-04-14
crashfix

olofkallander:
IceSession and transport keep the Tcp endpoints until the very end.
If new UDP PRFLX candidates show up after ICE connected, IceSession will be able to iterate through the endpoints to setup probe for the new candidate without crashing.

### DIFF
--- a/Jenkins/Jenkinsfile
+++ b/Jenkins/Jenkinsfile
@@ -29,7 +29,7 @@ node('be-integration') {
                 isReleaseBuild = true
             } else if (env.GIT_BRANCH ==~ /\d+\.\d+\.\d+/) {
                 version = "${env.GIT_BRANCH}-${currentBuild.number}"
-                isReleaseBuild = true
+                isReleaseBuild = !env.GIT_BRANCH.startsWith("0");
             }
 
             // Set the version env variable so that it's picked up by versioninfo.sh run as part of the build
--- a/transport/TransportImpl.cpp
+++ b/transport/TransportImpl.cpp
@@ -535,13 +535,13 @@ TransportImpl::TransportImpl(jobmanager::JobManager& jobmanager,
         {
             auto addressPort = iceConfig.publicIpv4;
             addressPort.setPort(endpoint->getLocalPort().getPort());
-            _rtpIceSession->addLocalCandidate(addressPort, endpoint.get());
+            _rtpIceSession->addLocalUdpCandidate(addressPort, endpoint.get());
         }
         if (!iceConfig.publicIpv6.empty() && endpoint->getLocalPort().getFamily() == AF_INET6)
         {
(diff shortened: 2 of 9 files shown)


Jenkins/Jenkinsfile
```diff
@@ -29,7 +29,7 @@ node('be-integration') {
                 isReleaseBuild = true
             } else if (env.GIT_BRANCH ==~ /\d+\.\d+\.\d+/) {
                 version = "${env.GIT_BRANCH}-${currentBuild.number}"
-                isReleaseBuild = true
+                isReleaseBuild = !env.GIT_BRANCH.startsWith("0");
```
olofkallander, line 32: This allows us to push branches like 0.0.x that are debug test branches automatically built and pushed to services artifactory. It is then possible to build and deploy mbr build from that.
Question: should we also automatically build debug if the isReleaseBuild is false?

transport/ice/IceSession.cpp
```diff
@@ -202,14 +202,21 @@ void IceSession::addProbeForRemoteCandidate(EndpointInfo& endpoint, const IceCan
 
     if (remoteCandidate.transportType == TransportType::UDP)
     {
-        IceCandidate& localCandidate =
-            *std::find_if(_localCandidates.begin(), _localCandidates.end(), [endpointAddress](const IceCandidate& c) {
+        auto localCandidateIt =
+            std::find_if(_localCandidates.begin(), _localCandidates.end(), [&endpointAddress](const IceCandidate& c) {
                 return endpointAddress == c.baseAddress && c.type == IceCandidate::Type::HOST;
             });
+        if (localCandidateIt == _localCandidates.end())
```
olofkallander, line 209: This is good protection but should not really ever happen.

transport/TransportImpl.cpp
```diff
@@ -1961,27 +1961,15 @@ void TransportImpl::onIceStateChanged(ice::IceSession* session, const ice::IceSe
...
 
-                logger::info("candidate selected %s %s, %s",
+                logger::info("candidate selected %s %s, %s-%s",
                     _loggableId.c_str(),
                     _peerRtpPort.getFamilyString().c_str(),
                     ice::toString(endpoint->getTransportType()).c_str(),
+                    ice::toString(candidatePair.first.type).c_str(),
                     ice::toString(candidatePair.second.type).c_str());
             }
         }
```
olofkallander, line 1972: This is the bug fix. This leads to deletion of TcpEndpoint and IceSession still has reference by raw pointer.
We could argue that it is better to remove it from IceSession and then delete it, but strictly we should keep the TCP connectio in case webRTC decides to send over it. One could also argue that is unlikely and wastes resources in SMB in terms of handles, memory and extra load on epoll.

RicardoMDomingues (approved):

PR #307, opened 2023-04-24
Ice crash fix and test

olofkallander:
Big PR but many changes are small and easily reviewable. Main areas to look into:
- IceSession fixes to avoid segv and stall in DTLS negotiation. Only real business logic change.
- Fake network modified to carry TCP tests and IPv6.
- Udp endpoint refactored not inheriting BaseUdpEndpoint so it can be used in fake network.
- Endpoint factory extended so we can support fake tcp endpoints.
- Tests rearranged and a few added.


bridge/Bridge.cpp
```diff
@@ -119,14 +119,25 @@ Bridge::~Bridge()
...
 }
 
 void Bridge::initialize(std::shared_ptr<transport::EndpointFactory> endpointFactory,
-    httpd::HttpDaemonFactory& httpdFactory)
+    httpd::HttpDaemonFactory& httpdFactory,
+    const std::vector<transport::SocketAddress>& interfaces)
 {
-    _localInterfaces = gatherInterfaces(_config);
-    if (_localInterfaces.size() == 0)
+    if (interfaces.empty())
```
olofkallander, line 131: This allows us to inject a fake interface list instead of having to open only 127.0.0.1 in emulation tests.

transport/TransportImpl.h
```diff
@@ -118,6 +118,7 @@ class TransportImpl : public RtcTransport,
     void setRemoteIce(const std::pair<std::string, std::string>& credentials,
         const ice::IceCandidates& candidates,
         memory::AudioPacketPoolAllocator& allocator) override;
+    void addRemoteIceCandidate(const ice::IceCandidate& candidate) override;
```
olofkallander, line 121: This allows client side to remote candidates later and have those probed. We do not use trickling from smb side, but it is used by clients and we need to test it somehow.

transport/TransportImpl.cpp
```diff
@@ -1969,17 +1989,28 @@ void TransportImpl::onIceStateChanged(ice::IceSession* session, const ice::IceSe
...
-            _rtpEndpoints.front()->unregisterListener(this);
-            _rtpEndpoints.erase(_rtpEndpoints.begin());
+            while (!_rtpEndpoints.empty() && _rtpEndpoints.back().get() != _selectedRtp &&
+                _rtpEndpoints.back()->getTransportType() == ice::TransportType::TCP)
+            {
+                logger::info("discarding %s, ref %ld",
+                    _loggableId.c_str(),
+                    _rtpEndpoints.back()->getName(),
+                    _rtpEndpoints.back().use_count());
+                _rtpIceSession->onTcpRemoved(_rtpEndpoints.back().get());
```
olofkallander, line 2001: This is the crash fix. To remove the endpoint from ice session before it is actually removed by transport factory.

transport/EndpointFactoryImpl.h
```diff
@@ -26,11 +31,38 @@ class EndpointFactoryImpl : public EndpointFactory
         RtcePoll& epoll,
         bool isShared);
 
-    virtual UdpEndpoint* createUdpEndpoint(jobmanager::JobManager& jobManager,
+    UdpEndpoint* createUdpEndpoint(jobmanager::JobManager& jobManager,
```
olofkallander, line 34: extend EndpointFactory to handle all endpoint creation. This allows testing with emulation of other endpoints than UDP.

test/integration/emulator/FakeRecordingEndpoint.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
+#include "transport/RecordingEndpoint.h"
+
+class FakeRecordingEndpoint : public transport::RecordingEndpoint
+{
+public:
+    FakeRecordingEndpoint() : _name("FakeRecordingEndpoint") {}
```
olofkallander, line 7: If we expand this a bit we could actually setup a recording sink in the fake network and verify that the data arrives as expected. We also get much better code coverage of the API and mixer setup.

transport/ice/IceSession.cpp
```diff
@@ -173,6 +176,11 @@ void IceSession::probeRemoteCandidates(const IceRole role, uint64_t timestamp)
     auto transmitTime = timestamp;
     for (auto* probe : _checklist)
     {
+        if (probe->state != CandidatePair::State::Waiting)
```
olofkallander, line 179: only trigger new probes.

transport/ice/IceSession.cpp
```diff
@@ -192,12 +200,28 @@ void IceSession::probeRemoteCandidates(const IceRole role, uint64_t timestamp)
...
 }
 
-void IceSession::addProbeForRemoteCandidate(EndpointInfo& endpoint, const IceCandidate& remoteCandidate)
+IceSession::CandidatePair* IceSession::addProbeForRemoteCandidate(EndpointInfo& endpoint,
+    const IceCandidate& remoteCandidate)
 {
     const auto endpointAddress = endpoint.endpoint->getLocalPort();
-    if (endpointAddress.getFamily() != remoteCandidate.address.getFamily())
+    if (endpointAddress.getFamily() != remoteCandidate.address.getFamily() ||
+        endpoint.endpoint->getTransportType() != remoteCandidate.transportType)
```
olofkallander, line 208: I noticed a bug here that did not really show in logs. If a new peer reflexive TCP candidate showed up we triggered probing from UDP endpoint towards the new candidate.

test/integration/emulator/TimeTurner.cpp
```diff
@@ -100,9 +100,22 @@ void TimeTurner::runFor(uint64_t durationNs)
          timestamp = getAbsoluteTime())
     {
         logger::awaitLogDrained(0.75);
-        if (!_sleeperCountdown.wait(1000))
+        // wait up to 1s for threads to go into nanoSleep
```
olofkallander, line 103: Sometimes test was stalled temporarily 1s here because a thread was doing ~JobQueue waiting for semaphore.
Now we wait 5ms and release some worker threads by advancing 64ns and thereby let the worker threads allow the waiting thread to continue.

reddvl1980 (approved):

RicardoMDomingues (approved):

PR #313, opened 2023-05-23
SRTP with SDES keys

olofkallander:
This enables SRTP using SDES keys in SMB.
Multiple modes can be requested in alloc request.
When transport is configured the decision is made to use either DTLS, SDES, null-cipher. It cannot be changed after that point.

### DIFF
--- a/api/AllocateEndpoint.h
+++ b/api/AllocateEndpoint.h
@@ -18,8 +18,8 @@ struct AllocateEndpoint
         bool ice;
         utils::Optional<bool> iceControlling;
 
-        bool dtls = false;
-        bool sdes = false;
+        bool dtls;
+        bool sdes;
     };
 
     struct Audio
--- a/api/AllocateEndpoint.h
+++ b/api/AllocateEndpoint.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "api/RtcDescriptors.h"
+#include "transport/dtls/SrtpProfiles.h"
 #include "utils/Optional.h"
 #include <cstdint>
 #include <string>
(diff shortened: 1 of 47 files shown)


api/AllocateEndpoint.h
```diff
@@ -11,12 +13,13 @@ struct AllocateEndpoint
...
     {
-        Transport() : ice(false), dtls(false) {}
+        Transport() : ice(false), dtls(false), sdes(false) {}
 
         bool ice;
         utils::Optional<bool> iceControlling;
 
-        bool dtls;
+        bool dtls = false;
+        bool sdes = false;
```
RicardoMDomingues, line 22: [minor] redundant initialization

api/Parser.cpp
```diff
@@ -178,6 +182,20 @@ api::Transport parsePatchEndpointTransport(const nlohmann::json& data)
         dtls.setup = dtlsJson["setup"].get<std::string>();
         transport.dtls.set(dtls);
     }
+    else if (data.find("sdes") != data.end())
+    {
+        const auto& sdesJson = data["sdes"];
+        srtp::AesKey aesKey;
+        const size_t decodedLength =
+            utils::Base64::decode(sdesJson["key"].get<std::string>().c_str(), aesKey.keySalt, sizeof(aesKey.keySalt));
+        aesKey.profile = api::utils::stringToSrtpProfile(sdesJson["profile"].get<std::string>());
```
RicardoMDomingues, line 191:
```suggestion
        aesKey.profile = api::utils::stringToSrtpProfile(getJsonFieldOrThrow(sdesJson, "profile").get<std::string>());
```

RicardoMDomingues (approved):

PR #316, opened 2023-05-25
Testcov vi

olofkallander:
Added some tests to increase code coverage.
I also found a bug or two in api request handler.
Test makes sure the different transport modes and encryption modes work together in all combinations.

### DIFF
--- a/bridge/MixerManager.cpp
+++ b/bridge/MixerManager.cpp
@@ -276,7 +276,7 @@ void MixerManager::stop()
             break;
     }
 
-    bool jobsDone = false;
+    std::atomic_bool jobsDone(false);
     _backgroundJobQueue.post([&jobsDone]() { jobsDone = true; });
     for (;; usleep(10000))
     {
--- a/bridge/MixerManager.cpp
+++ b/bridge/MixerManager.cpp
@@ -39,10 +39,14 @@ namespace bridge
 class MixerManagerMainJob : public jobmanager::MultiStepJob
 {
 public:
-    MixerManagerMainJob(MixerManager& mixerManager, std::atomic_bool& running, utils::Pacer& statsRefreshPacer)
+    MixerManagerMainJob(MixerManager& mixerManager,
+        std::atomic_bool& running,
+        utils::Pacer& statsRefreshPacer,
+        std::atomic_bool& maintenanceRunning)
         : _mixerManager(mixerManager),
           _running(running),
-          _statsPacer(statsRefreshPacer)
+          _statsPacer(statsRefreshPacer),
+          _maintenanceRunning(maintenanceRunning)
     {
         _statsPacer.tick(utils::Time::getAbsoluteTime());
     }
@@ -69,10 +73,13 @@ public:
         return _running.load();
     }
 
+    ~MixerManagerMainJob() { _maintenanceRunning = false; }
+
 private:
     MixerManager& _mixerManager;
     std::atomic_bool& _running;
     utils::Pacer& _statsPacer;
+    std::atomic_bool& _maintenanceRunning;
 };
 
 MixerManager::MixerManager(utils::IdGenerator& idGenerator,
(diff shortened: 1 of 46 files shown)


RicardoMDomingues (approved):

bridge/endpointActions/ConferenceActions.cpp
```diff
@@ -688,7 +688,7 @@ void configureVideoEndpoint(const api::EndpointDescription& endpointDescription,
         if (transport.dtls.isSet())
         {
             const auto& dtls = transport.dtls.get();
-            const bool isRemoteSideDtlsClient = dtls.isClient() == 0;
```
olofkallander, line 691: bug here that we do not encounter often I think.

bridge/MixerManager.cpp
```diff
@@ -276,6 +276,16 @@ void MixerManager::stop()
             break;
     }
 
+    bool jobsDone = false;
```
olofkallander, line 279: dead lock at test shutdown. I believe it can happen in production too if there is a race for removing Mixer object when Bridge is in ~Bridge.

transport/ice/IceSession.cpp
```diff
@@ -832,13 +832,13 @@ bool IceSession::isGatherComplete(uint64_t now)
...
 {
-    uint64_t oldestSrflxCandidate = 0;
+    int64_t oldestSrflxCandidate = 0;
     for (auto& candidatePair : _candidatePairs)
     {
         if (candidatePair->gatheringProbe && candidatePair->state == CandidatePair::Succeeded &&
             candidatePair->remoteCandidate.type == IceCandidate::Type::RELAY)
         {
-            oldestSrflxCandidate = std::max(oldestSrflxCandidate, now - candidatePair->startTime);
+            oldestSrflxCandidate = std::max(oldestSrflxCandidate, utils::Time::diff(candidatePair->startTime, now));
```
olofkallander, line 841: protection for negative timestamp

transport/ice/IceSession.cpp
```diff
@@ -873,13 +873,17 @@ bool IceSession::isIceComplete(const uint64_t now)
...
+    if (utils::Time::diffGT(_sessionStart, now, _config.connectTimeout * utils::Time::ms))
     {
         return true;
     }
 
-    if (failCount > 0 && failCount == _checklist.size())
+    // if all candidates failed, it could be that we just had unreachable HOST candidates. Give it some time for a PRFLX
+    // to show up.
+    if (failCount > 0 && failCount == _checklist.size() &&
+        probeAge > static_cast<int64_t>(_config.additionalCandidateTimeout * utils::Time::ms))
```
olofkallander, line 884: Without time check here thw prflx must arrive before the first probe (that is host and will fail) has failed.
Usually that does occure but if it is tcp or something it can fail fast.

transport/ice/IceSession.cpp
```diff
@@ -1103,6 +1107,11 @@ void IceSession::setRemoteCredentials(const std::pair<std::string, std::string>&
 {
     DBGCHECK_SINGLETHREADED(_mutexGuard);
     _credentials.remote = credentials;
+    if (credentials.second.empty())
```
olofkallander, line 1110: crash protection

transport/ice/IceSession.cpp
```diff
@@ -1456,13 +1457,13 @@ void IceSession::CandidatePair::processTimeout(const uint64_t now)
     if (state == InProgress)
     {
         if (remoteCandidate.type == IceCandidate::Type::HOST &&
-            now - startTime > _config.hostProbeTimeout * utils::Time::ms)
+            utils::Time::diffGT(startTime, now, _config.hostProbeTimeout * utils::Time::ms))
```
olofkallander, line 1460: This could happen due to ICE request with prflx candidate arriving. Timestamp is then from transport which may be timstamped before the HOST probe started from SMB. The probe would fail immediately and the entire ice session failed since there are no other canddiates yet.

transport/dtls/SrtpClient.cpp
```diff
@@ -847,6 +847,12 @@ void SrtpClient::setRemoteKey(const srtp::AesKey& key)
     {
         _state = State::CONNECTED;
         _mode = srtp::Mode::NULL_CIPHER;
+        logger::debug("SRTP connected with null cipher", _loggableId.c_str());
+        if (_eventSink)
+        {
+            _eventSink->onDtlsStateChange(this, _state);
```
olofkallander, line 853: Maybe this should be called onSrtpStateChange becasue it has to be signalled also in null cipher mode and SDES that we are connected on SRTP level.

PR #322, opened 2023-08-15
fix ROC

olofkallander:
SRTP requires ROC to be set after a longer seqno gap, not only when crossing wrapping point. Otherwise, it may not be able to decrypt. Eg. decrypt 12, 82630, 115999. The 3rd will fail though it has same ROC as 2nd and we set ROC before 2nd.
Fix is to set ROC on 15 bit wrap.

SRTP context must see at least one packet with ROC=0. Discarded video streams pose a problem and now we always decipher one packet with ROC=0

### DIFF
--- a/bridge/engine/AudioForwarderReceiveJob.cpp
+++ b/bridge/engine/AudioForwarderReceiveJob.cpp
@@ -149,13 +149,6 @@ bool AudioForwarderReceiveJob::unprotect(memory::Packet& opusPacket)
 
     if (!_sender->unprotect(opusPacket))
     {
-        logger::error("Failed to unprotect srtp %u, extseqno %u->%u, %s, %s",
-            "AudioForwarderReceiveJob",
-            _ssrcContext.ssrc,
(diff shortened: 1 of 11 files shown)


bridge/engine/AudioForwarderReceiveJob.cpp
```diff
@@ -130,7 +132,7 @@ bool AudioForwarderReceiveJob::unprotect(memory::Packet& opusPacket)
             oldRolloverCounter,
             newRolloverCounter,
             _sender->getLoggableId().c_str());
-        if (!_sender->setSrtpRemoteRolloverCounter(_ssrcContext.ssrc, newRolloverCounter))
+        if (!_sender->setSrtpRemoteRolloverCounter(_ssrcContext.ssrc, _extendedSequenceNumber >> 16))
```
reddvl1980, line 135: Why not 'newRolloverCounter'?

reddvl1980 (approved):

PR #325, opened 2023-09-19
audio receive pipe line for mix mode

olofkallander:
This is a first commit that does not change how SMB performs mixing. 
It is unit tests for an audio receive pipe line that handles incoming jitter, audio decoding, time compression in concert.


codec/AudioTools.cpp
```diff
@@ -0,0 +1,206 @@
...
+        {
+            ++i;
+        }
+
+        writeIndex += 2;
+    }
+    return writeIndex / 2;
+}
+
+void sineTail(int16_t* data, double freq, uint32_t sampleRate, size_t count)
```
olofkallander, line 38: can be removed before merge if we are happy with the swingTail output

rtp/RtpDelayTracker.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    if (_renderTime == 0 && _rtpTimestamp == 0)
+    {
+        _renderTime = receiveTime;
+        _rtpTimestamp = rtpTimestamp;
+        _delay = 0;
+        return 0;
+    }
+
+    _renderTime += static_cast<int32_t>(rtpTimestamp - _rtpTimestamp) * static_cast<int64_t>(utils::Time::sec) /
```
RicardoMDomingues, line 26: This is calculating the renderTime with slight lower frequency. So if the real frequency is correct or higher, that will increasing the reported delay over time, right? I don't find a point where this can be fixed for more real values in that case. How that affects the final jitter measurement?

PR #326, opened 2023-09-27
MEDIA-2058: log 'ERROR' level messages into stderr as well (when log …

reddvl1980:
…to stdout is enabled)

### DIFF
--- a/logger/LoggerThread.cpp
+++ b/logger/LoggerThread.cpp
@@ -58,6 +58,9 @@ void LoggerThread::run()
             if (_logStdOut)
             {
                 formatTo(stdout, localTime, item->logLevel, item->threadId, item->message);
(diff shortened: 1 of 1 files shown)


logger/LoggerThread.cpp
```diff
@@ -58,6 +58,9 @@ void LoggerThread::run()
             if (_logStdOut)
             {
                 formatTo(stdout, localTime, item->logLevel, item->threadId, item->message);
+                if (!(std::strcmp(item->logLevel, "ERROR"))) {
```
olofkallander, line 61: ( after ! is not needed

olofkallander (approved):

logger/LoggerThread.cpp
reddvl1980: fixed.

PR #327, opened 2023-09-29
Alternative FFT for test

olofkallander:
merge fft spectrum

### DIFF
--- a/bridge/engine/EngineMixerAudio.cpp
+++ b/bridge/engine/EngineMixerAudio.cpp
@@ -461,7 +461,8 @@ void EngineMixer::processAudioStreams()
             for (auto& stream : _engineAudioStreams)
             {
                 auto& peerAudioStream = *stream.second;
-                if (peerAudioStream.remoteSsrc.isSet() && peerAudioStream.transport.isConnected() &&
+                if (stream.second->endpointIdHash != audioStream->endpointIdHash &&
+                    peerAudioStream.remoteSsrc.isSet() && peerAudioStream.transport.isConnected() &&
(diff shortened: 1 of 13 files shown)


bridge/engine/EngineMixerAudio.cpp
```diff
@@ -461,7 +461,8 @@ void EngineMixer::processAudioStreams()
             for (auto& stream : _engineAudioStreams)
             {
                 auto& peerAudioStream = *stream.second;
-                if (peerAudioStream.remoteSsrc.isSet() && peerAudioStream.transport.isConnected() &&
+                if (stream.second->endpointIdHash != audioStream->endpointIdHash &&
```
olofkallander, line 464: This was a bug slip through. Mix stream is already subtracted because it is contributing, but since it is also a neighbour of itself it is subtracted twice. The analyzeRec did not pick up the frequency because the audio level is different for mix stream.

test/integration/SampleDataUtils.cpp
```diff
@@ -360,6 +373,69 @@ void SampleDataUtils::ifft(SampleDataUtils::CmplxArray& x)
...
+    double f = 0;
+    for (auto m : powerSpectrum)
+    {
+        v.push_back({f, m});
+        f += freqDelta;
+    }
+    return v;
+}
+
+std::vector<std::pair<double, double>> SampleDataUtils::isolatePeaks(std::vector<std::pair<double, double>>& powerFreq,
```
olofkallander, line 422: Perhaps we will use this more in tests. It tuned out to be more efficient for c9 where the tone is poluted.

PR #336, opened 2023-11-03
Emailext

olofkallander:
test email ext
It would be nice if we could make this work as it is easier for contributors to address test problems that is not visible to them.
Right now it executes the send but nothing arrives. Either the serivce account does not allow it, the attachment is wrong or the email address is not what we think it is.

### DIFF
--- a/Jenkins/PRUnitTestRunner.groovy
+++ b/Jenkins/PRUnitTestRunner.groovy
@@ -26,11 +26,9 @@ void prRunner(String cmakeBuildType, String platform, String dockerTag) {
                 sh "docker/$platform/runtests.sh"
             }
         }
-//#recipientProviders: [[$class: 'CulpritsRecipientProvider']]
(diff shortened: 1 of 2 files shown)


Jenkins/PRUnitTestRunner.groovy
```diff
@@ -25,6 +25,23 @@ void prRunner(String cmakeBuildType, String platform, String dockerTag) {
             sh "docker/$platform/runtests.sh"
         }
     }
+//#recipientProviders: [[$class: 'CulpritsRecipientProvider']]
+// maybe remove to
+    post {
```
RicardoMDomingues, line 30: Is this works with scripted pipelines? This seems like declarative pipeline syntax

olofkallander: could be that is why it is not working.

PR #337, opened 2023-11-03
RTC-14634

olofkallander:
Automatic fail over to another candidate when current one becomes unresponsive.

also fix for keeping conference alive on ICE activity on the transports.

### DIFF
--- a/transport/ice/IceSession.cpp
+++ b/transport/ice/IceSession.cpp
@@ -19,6 +19,8 @@ IceSession::IceSession(size_t sessionId,
     const IceRole role,
     IEvents* eventSink)
     : _logId("IceSession-" + std::to_string(sessionId)),
+      _nomination(nullptr),
+      _preliminaryCandidate(nullptr),
       _component(component),
       _tcpProbeCount(0),
       _config(config),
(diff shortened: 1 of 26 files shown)


reddvl1980 (approved):

transport/ice/IceSession.cpp
```diff
@@ -984,22 +1050,50 @@ void IceSession::stateCheck(const uint64_t now)
...
     }
 }
 
 void IceSession::freezePendingProbes(uint64_t now)
 {
     for (auto& candidatePair : _candidatePairs)
     {
-        if (!candidatePair->nominated && candidatePair->state < CandidatePair::State::Failed &&
-            candidatePair->localCandidate.transportType == ice::TransportType::TCP)
+        if (!candidatePair->nominated && candidatePair->state <= CandidatePair::State::Succeeded)
```
RicardoMDomingues, line 1078: Why did you remove the `candidatePair->localCandidate.transportType == ice::TransportType::TCP` condition?

I think probes are like a keep-alive, if we don't probe them, NAT gateways can close them potentially by inactivity and then if we need to re-probe they might fail.

PR #338, opened 2023-11-04
cholesky update

olofkallander:
Added more tests around cholesky decomposition. 
Also make sure the covariance matrix is symmetric before it is stored in bwe. If it is not symmetric there is a slight risk cholesky decomposition fails and the entire estimation collapses.


RicardoMDomingues (approved):

test/bwe/MatrixTests.cpp
```diff
@@ -46,9 +53,154 @@ TEST(MatrixTest, choleskyStability)
...
+    auto d = m - c;
+    auto nrm = math::norm(m - c);
+    EXPECT_LT(math::norm(m - c), 1.E-10);
     EXPECT_TRUE(math::isValid(L));
     // should check determinant of m - c
-}
\ No newline at end of file
+}
+
+#define EXPECT_ERROR_LT(a, b, relativeError)                                                                           \
```
RicardoMDomingues, line 75: This is a re-implementation of `ASSERT_NEAR ` provided by gtest

https://github.com/google/googletest/blob/main/docs/reference/assertions.md#expect_double_eq-expect_double_eq

test/bwe/MatrixTests.cpp
```diff
@@ -46,9 +53,154 @@ TEST(MatrixTest, choleskyStability)
...
+        randomize(m, 10000.0);
+
+        const auto d1 = det2(m);
+        const auto d2 = det2(transpose(m));
+        const auto d3 = det(m);
+        const auto d4 = det(transpose(m));
+
+        if (d2 != d4)
+        {
+            if (det(m) == 0 || det(transpose(m)) == 0)
```
RicardoMDomingues, line 103: this `if` block is doing nothing

test/bwe/MatrixTests.cpp
olofkallander: EXPECT_NEAR is comparing with absolute error. This one compares relative error.
It can be done differently I guess.

PR #352, opened 2024-01-26
MEDIA-2265: Change the api in order to support Named Telephone Events

RicardoMDomingues:
This is proposal to change API in order to support  Named Telephone Events.

Currently API contains a single json object to describe the audio payload. With this change, we are making it an array of payloads instead, in the similar way we do for video payloads. The json property was then renamed from `payload-type` to `payload-types`.

If you need to support multiple SMB versions, then your code should be prepared to encounter `payload-type` as a single object xor `payload-types` as an array of objects when you allocate an endpoint, pseudo-code example:

```
if (audioJsonNode.has("payload-type")) {
    // old version. Process as you did before
    addPayloadToMediaDescription(audioJsonNode.get("payload-type"));
} else if (audioJsonNode.has("payload-types"))
    // New version
     auto payloadsArray = audioJsonNode.get("payload-types");
     for (auto payloadJsonObject : payloadsArray) {
          addPayloadToMediaDescription(payloadJsonObject);
      }
}
``` 

If you are supporting multiple SMB versions, then you need to keep in mind to use `payload-type` or `payload-types` in configuration and reconfiguration actions, depending on which SMB version you are calling

## Main Modifications
 - **Refactor AudioRewrite and VideoRewrite** - Centralize the logic in `SsrcOutboundContext` as the code is similar. Also `SsrcOutboundContext` it was already deciding if a packet should be forwarded or not and the rewrite was supporting (tested by unit tests) unrealistic scnarios where the packet it would be dropped earlier. The method `shouldSend` was deleted and `rewriteAudio` and `rewriteVideo` returns false when the packet should not be sent!
 - **Implement the forward logic for telephone events** - Telephone events are received in a different payload type but in the same ssrc and sequence of audio packets. Then we can't have a specific queues for telephone events to them not get reordered! It there still is a limitation on telephone events over barbells. We will **not support it for now**


marcusspangenberg:
> Hey @marcusspangenberg I know you are/were using SMB. I am doing this breaking change on API to support Named Telephone Events. While this requires code changes on components that communicates with SMB, I believe the changes should be pretty straight forward and it should be relatively easy if you need to support multiple SMB versions.
> 
> Please let me know if you might encounter any obstacles with this change in your project.

Thanks for the mention @RicardoMDomingues. Much appreciated. The change won't cause any problems for our use case.

bridge/engine/SsrcOutboundContext.cpp
```diff
@@ -1,33 +1,437 @@
...
+    assert(!(isTelephoneEvent && telephoneEventRtpMap.isEmpty()));
+
+    header.sequenceNumber = newSequenceNumber & 0xFFFFu;
+    header.ssrc = this->ssrc;
+    header.timestamp = _rewrite.offset.timestamp + header.timestamp;
+    header.payloadType = isTelephoneEvent ? telephoneEventRtpMap.payloadType : rtpMap.payloadType;
+    doRtpHeaderExtensionRewriteForAudio(header, senderInboundContext);
+}
 
-bool SsrcOutboundContext::shouldSend(uint32_t ssrc, uint32_t sequenceNumber) const
```
RicardoMDomingues, line 6: Deciding if the packet should be sent is now part of the rewrite code logic itself. All rewrite code was centralized in `SssrcOuboundContext` and the logic about whether  the rewrite is possible or not should not be scattered over multiple files as it makes the logic error prone

reddvl1980 (approved):

PR #354, opened 2024-02-02
MEDIA-2365: Send user map periodic over barbell

RicardoMDomingues:
On C9 we have scenario of recovering where sfumanager can lose track of a barbell and a new barbell will be set up. For this to happen, there is 1 leg that is going to be removed but in 1 region there will be a barbell leaked until the end of the conference.

On this side where the barbell leg leaked, the barbell is o going to be deactivated by timeout and all ssrc inbound context will be decommissioned, when that time we have already a new barbell to replace this broken one, the decommission process can remove the participant endpoint id from the user map. To mitigate this issue, this commit allow us to configure an interval where the user mapping is going to be send over barbells periodically to recovery the problem caused by decommission the old ssrc context.

This is just a safety mechanism as in C9 setting up barbell is super complex with split ownership on the 2 regions involved, for the Symphony use case this setting shouldn't be necessary

### DIFF
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -55,8 +55,10 @@ EngineMixer::EngineMixer(const std::string& id,
       _mainAllocator(mainAllocator),
       _sendAllocator(sendAllocator),
       _audioAllocator(audioAllocator),
-      _lastReceiveTimeOnRegularTransports(utils::Time::getAbsoluteTime()),
-      _lastReceiveTimeOnBarbellTransports(utils::Time::getAbsoluteTime()),
+      _lastStartedIterationTimestamp(utils::Time::getAbsoluteTime()),
+      _lastReceiveTimeOnRegularTransports(_lastStartedIterationTimestamp),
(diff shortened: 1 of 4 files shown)


bridge/engine/EngineMixerBarbell.cpp
```diff
@@ -17,6 +17,85 @@
 namespace bridge
 {
 
+namespace
```
RicardoMDomingues, line 20: Put the anonymous namespace at the top

bridge/engine/EngineMixer.cpp
```diff
@@ -1318,6 +1322,12 @@ void EngineMixer::sendMessagesToNewDataStreams()
             continue;
         }
 
+        if (!isDominantSpeakerMessageBuild)
+        {
+            _activeMediaList->makeDominantSpeakerMessage(dominantSpeakerMessage);
```
RicardoMDomingues, line 1327: Minor improvement,  build a message only there is at least 1 participant that should receive it

reddvl1980 (approved):

PR #359, opened 2024-03-08
Rebase/pcma

RicardoMDomingues:


olofkallander (commented):
Actually I have a much better G711 handling in another branch. It does not rely on speex lib, and is more efficient in resampling for example.

RicardoMDomingues:
Closed in favor of #365

PR #364, opened 2024-04-04
Ensure Ensure 8 byte alignment

RicardoMDomingues:


RicardoMDomingues:
Close as it seems alignments are ok

PR #370, opened 2024-05-05
Bwe4g

olofkallander:
Improved performance on LTE network. Bw estimator identifies true bandwidth. Bwe converges accurately on WiFi and handles low bandwidth more accurately. In theory it can measure low bandwidth in audio only mode. In practice this may work for links with low jitter and low bandwidth. Also, the mobile devices webRTC does not seem to add audio send timestamps, so there it will not work at all.


bwe/BandwidthEstimator.cpp
```diff
@@ -31,164 +43,243 @@ BandwidthEstimator::BandwidthEstimator(const Config& config)
...
+        actualDelay = 0;
+        _packetSize0 = packetSize;
+    }
+    else if (actualDelay == 0 && packetSize > _packetSize0)
+    {
+        _packetSize0 = packetSize;
+    }
+    else if (observedDelay > expectedDelay && tau > 20.0 && expectedState(QueuedBits) < _config.mtu * 8 * 3)
+    {
+        // inter transmit interval larger than 20ms risk having radio shut down. adjust co up
```
olofkallander, line 116: For LTE network, the radio can be put to sleep when there is a transmission pause. It does not care that it is running behind. This allows bwe to measure observed throughput. Improvement is +1Mbps.
It does not allow major increase, which could lead to over estimation after pause on back bone link. It merly makes the first packet fit current estimate. If this is wrong assumption the clock offset will quickly be adjusted down. Note that we require queue to be drained, and that is when this matters most.

bwe/BandwidthEstimator.cpp
```diff
@@ -31,164 +43,243 @@ BandwidthEstimator::BandwidthEstimator(const Config& config)
...
+    // calculate where we would have transitioned from alternative positions, and what the delay would be then
     std::array<double, SIGMA_POINTS> predictedDelays;
     for (size_t i = 0; i < sigmaPoints.size(); ++i)
     {
-        sigmaPoints[i] = transitionState(packetSize, tau, sigmaPoints[i], observedDelay);
-        predictedDelays[i] = predictDelay(sigmaPoints[i]);
+        sigmaPoints[i] = transitionState(packetSize, tau, sigmaPoints[i]);
+        predictedDelays[i] = predictAbsoluteDelay(sigmaPoints[i]);
     }
+    const double predictedMeanDelay = predictedDelays[0]; // delay of mean state, because 1/bw is non-linear
```
olofkallander, line 146: I noticed that the mean delay from all sigma points is always offset upwards, because the transformation is non-linear. 
Using the delay from the current state is a better starting point for the variance calculation, though it bends ukf a bit.

bwe/BandwidthEstimator.cpp
```diff
@@ -31,164 +43,243 @@ BandwidthEstimator::BandwidthEstimator(const Config& config)
...
     for (size_t i = 1; i < sigmaPoints.size(); ++i)
     {
         statePredictionCovariance += math::outerProduct(sigmaPoints[i]);
     }
-    statePredictionCovariance *= _weightMeanCovariance;
+    statePredictionCovariance *= _weightCovariance;
     statePredictionCovariance += (_weightCovariance0 * math::outerProduct(sigmaPoints[0]));
     assert(math::isValid(statePredictionCovariance));
 
-    double predictedMeanDelay = predictedDelays[0] * _weightMean0;
```
olofkallander, line 115: we take mean from current state instead.

bwe/BandwidthEstimator.cpp
```diff
@@ -31,164 +43,243 @@ BandwidthEstimator::BandwidthEstimator(const Config& config)
...
     }
-    crossCovariance *= _weightMeanCovariance;
+    crossCovariance *= _weightCovariance;
     crossCovariance += _weightCovariance0 * residual0 * sigmaPoints[0];
 
+    // update position towards mean position
     const auto kalmanGain = crossCovariance * (1.0 / covDelay);
     _state = predictedMeanState + kalmanGain * (observedDelay - predictedMeanDelay);
-    _covarianceP = statePredictionCovariance - kalmanGain * covDelay * math::transpose(kalmanGain);
+    _covarianceP = statePredictionCovariance - crossCovariance * math::transpose(kalmanGain);
```
olofkallander, line 189: Equivalent to above. One less vector multiplication

bwe/BandwidthEstimator.cpp
```diff
@@ -31,164 +43,243 @@ BandwidthEstimator::BandwidthEstimator(const Config& config)
...
+    state(QueuedBits) = std::max(packetBits, state(QueuedBits));
+
+    if (observedDelay - predictAbsoluteDelay(state) < 0 && state(QueuedBits) > _config.mtu * 3)
     {
-        // under estimating delay. Likely in congestion and over estimating bw
-        _congestion.margin = _config.congestion.backOff;
-        _state(1) = std::min(_state(1), _receiveBandwidth.get(550 * utils::Time::ms) * utils::Time::ms);
+        // adjust queue if we received packet earlier than expected. If bw is higher than expected, queue should also
+        // have drained more
+        const double delayErr = predictAbsoluteDelay(state) - observedDelay;
```
olofkallander, line 216: This avoids bw over estimation when bw goes up. The model queue is always lagging behind as it is modeled on previous lower drain rate.

bwe/BandwidthEstimator.cpp
```diff
@@ -31,164 +43,243 @@ BandwidthEstimator::BandwidthEstimator(const Config& config)
...
+
+    const double longerQueue = _config.mtu * 8 * 2;
+    if (_congestion.consecutiveOver == 0)
+    {
+        _congestion.estimateBeforeCongestion = currentState(Bandwidth);
+    }
+
+    if (_congestion.consecutiveUnder > 15 || _congestion.consecutiveOver > 30)
+    {
+        // adapt faster at consistent observed low delay or prolonged observed higher delay
```
olofkallander, line 244: If we observe lower delay consistently we should look for lower bandwidth and adjust faster to observation.
If we observe higher delay than expected we wait a bit longer before adjusting bw.
We adjust more, the longer this has been observed.

bwe/BandwidthEstimator.cpp
```diff
@@ -221,28 +312,37 @@ double BandwidthEstimator::analyseCongestion(double actualDelay, const uint32_t
...
     return congestionScale;
 }
 
 // in kbps
 double BandwidthEstimator::getEstimate(uint64_t timestamp) const
 {
-    const double estimatedBandwidth = std::min(_state(1), _congestion.dip.bandwidthCapKbps);
+    double estimatedBandwidth = std::min(_state(Bandwidth), _congestion.dip.bandwidthCapKbps);
+    if (_congestion.consecutiveOver < 50)
+    {
```
olofkallander, line 324: If we have congestion for some time, we do not immediately drop bw as this problem may go away (LTE). We keep the estimate as observed before congestion for 50pkt. (Usually about 0.5s). If it does not improve by then we let it drop to the model estimate.

bwe/BandwidthEstimator.cpp
```diff
@@ -221,28 +312,37 @@ double BandwidthEstimator::analyseCongestion(double actualDelay, const uint32_t
...
+        return math::clamp(estimatedBandwidth * (1.0 - _config.silence.backOff),
+            _config.estimate.minReportedKbps,
+            _config.silence.maxBandwidthKbps);
     }
-    return std::max(std::max(_congestion.dip.bandwidthFloorKbps, _config.estimate.minReportedKbps),
-        estimatedBandwidth * (1.0 - _congestion.margin));
+
+    return std::max({_congestion.dip.bandwidthFloorKbps,
+        _config.estimate.minReportedKbps,
+        estimatedBandwidth * (1.0 - _congestion.margin)});
```
olofkallander, line 339: Congestion margin is used for severe choking and we try to cause sender to allow for link to drain in less than 10s.
Otherwise, estimator will continue to report full bandwidth and delay will persist for long time.

bwe/BandwidthEstimator.cpp
```diff
@@ -255,94 +355,90 @@ math::Matrix<double, 3> BandwidthEstimator::getCovariance() const
...
+    sigmaPoints[0] = state;
 
     int startIndex = 1;
+    const double maxBandwidthDeviation = std::max(0.0, state(Bandwidth) - 10);
     for (int c = 0; c < squareRoot.columns(); ++c)
     {
         auto sigmaOffset = _sigmaWeight * squareRoot.getColumn(c);
-        sigmaPoints[startIndex + 2 * c] = _state + sigmaOffset;
-        sigmaPoints[startIndex + 2 * c + 1] = _state - sigmaOffset;
+        sigmaOffset(QueuedBits) = math::clamp(sigmaOffset(QueuedBits), -state(QueuedBits), state(QueuedBits));
```
olofkallander, line 374: This is important to keep the mean state centered on the current state queue and not allow negative queue in hypothetical states.

bwe/BandwidthEstimator.cpp
```diff
@@ -255,94 +355,90 @@ math::Matrix<double, 3> BandwidthEstimator::getCovariance() const
...
+    for (int i = 0; i < processNoise.rows(); ++i)
     {
         math::Matrix<double, 3> noise;
         noise(i) = processNoise(i) * _sigmaWeight;
-        sigmaPoints[startIndex + i * 2] = _state + noise;
-        // Offset the positive clock offset sigma point. Better catch up to clock drift.
-        // CO is easily pushed down by packets with short delay.
-        sigmaPoints[startIndex + i * 2 + 1] = _state - (i == 2 ? noise * 0.001 : noise);
+        noise(QueuedBits) = math::clamp(noise(QueuedBits), -state(QueuedBits), state(QueuedBits));
+        noise(Bandwidth) = math::clamp(noise(Bandwidth), -maxBandwidthDeviation, maxBandwidthDeviation);
```
olofkallander, line 387: This is important for symetric bw hypothesis around current bw. Otherwise, the bw will drift off.

bwe/BandwidthEstimator.cpp
```diff
@@ -255,94 +355,90 @@ math::Matrix<double, 3> BandwidthEstimator::getCovariance() const
...
     startIndex += 2 * squareRoot.columns();
-    for (int i = 0; i < _processNoise.rows(); ++i)
+    for (int i = 0; i < processNoise.rows(); ++i)
     {
         math::Matrix<double, 3> noise;
         noise(i) = processNoise(i) * _sigmaWeight;
-        sigmaPoints[startIndex + i * 2] = _state + noise;
-        // Offset the positive clock offset sigma point. Better catch up to clock drift.
-        // CO is easily pushed down by packets with short delay.
-        sigmaPoints[startIndex + i * 2 + 1] = _state - (i == 2 ? noise * 0.001 : noise);
```
olofkallander, line 283: Tests show this is no longer needed. It was a problem caused by the non-linear transform. Clock offset accurately follows the clock drift as observed in real traces.

bwe/BandwidthEstimator.cpp
```diff
@@ -255,94 +355,90 @@ math::Matrix<double, 3> BandwidthEstimator::getCovariance() const
...
 
     // add two points for measurement noise
-    startIndex += 2 * _processNoise.rows();
+    startIndex += 2 * processNoise.rows();
     sigmaPoints[startIndex] = sigmaPoints[0];
     sigmaPoints[startIndex + 1] = sigmaPoints[0];
-
-    for (auto& point : sigmaPoints)
-    {
-        point(0) = std::max(0.0, point(0)); // no magic Q reduction
```
olofkallander, line 293: assured by the sigma point offsets instead.

test/bwe/FakeAudioSource.cpp
```diff
@@ -74,6 +74,6 @@ size_t FakeAudioSource::randomPacketSize()
         }
     }
     ++_counter;
-    return _talkSprint + randomSize(_talkSprint, 0.1);
+    return _talkSprint + randomSize(_talkSprint / 4, 0.75);
```
olofkallander, line 77: This caused a bit excessive bitrate beyond allocation.

test/bwe/FakeVideoSource.cpp
```diff
@@ -151,14 +164,29 @@ memory::UniquePacket FakeVideoSource::getPacket(uint64_t timestamp)
...
     ++_counter;
     _packetsInFrame = 0;
     _frameSize = randomSize(meanSize, 0.2);
     _pacing = (utils::Time::sec / _fps) * _mtu / (2 * (_frameSize + _mtu));
 }
+
+void FakeVideoSource::setBandwidth(uint32_t kbps)
+{
+    _bandwidthKbps = kbps;
+    _fps = 30;
```
olofkallander, line 182: emulate reduction of fps on lower bandwidths.

test/integration/emulator/TimeTurner.cpp
```diff
@@ -160,7 +165,7 @@ void TimeTurner::advance()
 
 void TimeTurner::advance(uint64_t nanoSeconds)
 {
-    nanoSeconds = std::max(nanoSeconds, 2 * utils::Time::ms);
+    nanoSeconds = std::max(nanoSeconds, _granularity);
```
olofkallander, line 168: granularity allows timer source o advance in smaller steps. I noticed this caused physically impossible behaviour on low bw links in emulation.

olofkallander:
test

olofkallander:
test

RicardoMDomingues (approved):

PR #373, opened 2024-06-11
Add missing size multiplier when zeroing mixedData

marcusspangenberg:
Working on gcc compiler support and I got a warning here. Seems like a bug.

### DIFF
--- a/bridge/engine/EngineMixer.cpp
+++ b/bridge/engine/EngineMixer.cpp
@@ -84,7 +84,7 @@ EngineMixer::EngineMixer(const std::string& id,
     assert(audioSsrcs.size() <= SsrcRewrite::ssrcArraySize);
     assert(videoSsrcs.size() <= SsrcRewrite::ssrcArraySize);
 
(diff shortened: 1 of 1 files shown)


bridge/engine/EngineMixer.cpp
```diff
@@ -84,7 +84,7 @@ EngineMixer::EngineMixer(const std::string& id,
     assert(audioSsrcs.size() <= SsrcRewrite::ssrcArraySize);
     assert(videoSsrcs.size() <= SsrcRewrite::ssrcArraySize);
 
-    std::memset(_mixedData, 0, samplesPerFrame20ms * channelsPerFrame);
+    std::memset(_mixedData, 0, samplesPerFrame20ms * channelsPerFrame * sizeof(int16_t));
```
RicardoMDomingues, line 87: I think it is safer this way:

```suggestion
    std::memset(_mixedData, 0, sizeof(_mixedData));
```

RicardoMDomingues (approved):

PR #375, opened 2024-06-18
OpenSSL 3 support

marcusspangenberg:
Add support for both OpenSSL 3 and 1.1.1 in related classes.

### DIFF
--- a/crypto/SslHelper.h
+++ b/crypto/SslHelper.h
@@ -2,11 +2,11 @@
 
 #include <cstddef>
 #include <cstdint>
-#include <memory>
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
 #include <openssl/opensslv.h>
 #include <string>
+#include <vector>
 
 namespace crypto
 {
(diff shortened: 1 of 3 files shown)


crypto/SslHelper.h
```diff
@@ -32,7 +34,18 @@ class HMAC
...
 
 private:
+#if OPENSSL_VERSION_MAJOR >= 3
+    EVP_MAC* _mac = nullptr;
+    EVP_MAC_CTX* _ctx;
+#else
     hmac_ctx_st* _ctx;
+#endif
+    std::unique_ptr<uint8_t[]> _key;
+    int32_t _keyLength;
```
RicardoMDomingues, line 44: You could replace this 2 fields by a std::vector.
It is already a byte array allocated on heap memory and you its easier to use than deal with nullptr and size in a separated variable

RicardoMDomingues (approved):
Just minor comment, otherwise looks good to me

RicardoMDomingues:
Fail in continuous-integration/jenkins/pr-head is not related to your change. It should be good to merge

transport/dtls/SslDtls.cpp
```diff
@@ -26,6 +26,9 @@ int verify(int, X509_STORE_CTX*)
 
 EVP_PKEY* generateRsaKey()
 {
+#if OPENSSL_VERSION_MAJOR >= 3
+    return EVP_RSA_gen(keySize);
+#else
```
olofkallander, line 31: I think actually this works already on openssl 1.1.1. The compiler compains about deprecated methods.

PR #377, opened 2024-06-20
MEDIA-2661: ICE reconnects during meeting and fails

RicardoMDomingues:
### Problem
We've had some issues with users complaining that they can't hear anyone, but are heard by others. After analyzing the logs we noticed that there are some problems when SMB tries to nominate a new ICE candidate that sometimes fails and let the ICE state in FAILED on SMB,  while clients have a different ICE state which cause the clients to continue in the meeting without trigger the recover (in Symphony this process creates a completely new session that takes over the older one).

We observed that in one of the cases, client presented 1 IPv6 candidate and 1 IPv4. SMB nominates IPv6 and discard IPv4 PRFLX candidate after CONNECTED state. Then something happened with IPv6 and SMB changed the ICE state to CONNECTING trying to select a new candidate and it failed after 15 seconds. Although it was not nominated a new candidate, client logs shows that WebRTC decided to use IPv4 candidate to send as the IPv6 was not reachable. The problem on IPv6 seems to had been a temporary glitch and clients logs shows that the client continued to receive media from SMB through the IPv6 after this event (but SMB was already on FAILED state), that works for a few minutes until it stops to work (perhaps client firewall close the port due to lack of outbound traffic through the IPv6) and client stops to receive media while it continued to send through the IPv4.

This is caused by this issues:
1. SMB stills to respond to client ICE requests when the state is FAILED. This could mean that clients do not fail the ICE on their side, leaving the SMB and clients with different views of the ICE state.
2. After ICE state goes to CONNECTED, SMB discards good PRFLX candidates that could potential be used as backup if something happens with the nominated.
3. After 15 seconds (`reflexiveProbeTimeout`) of being in CONNECTED state, SMB stops to accept new PRFLX candidates. SMB can accept PRFLX candidates again if it goes to CONNECTING state but it has only 15 seconds to see the PRFLX candidates before fail. WebRTC client seems to be quick on detection of ICE problems and they can start probe all backups candidates again before SMB goes to CONNECTING and spend the next 25 seconds without probing them again, causing SMB to fail before see the PRFLX candidates.

### Solution
With this change we stop to respond to ICE STUN request when the ICE state on SMB is FAILED. SMB does not have any mechanism to report the ICE state to the components that manage the signalling with clients nor it has a way to restart the ICE (which involves a new SDP offer-answer negation with a different ICE username). Then, the best decision SMB can take is stop to respond to ICE STUN request which will cause the ICE state on client to go to FAILED as well and then clients can trigger the recover on their end.

Also we don't discard good PRFLX candidates only because they were not nominated. They will be kept as backup candidates if the nominated becomes unreachable. This relies on the fact that the main WebRTC implementation (Google implementation) keeps probing the backup candidates each 25 seconds, preventing to NAT/firewall to recycle the port while SMB has the candidates pair on frozen state.

We also accept candidates anytime. This is not required as the initial problem was discarding backup candidates. So if we keep the backup candidates, then we could stop accept new candidates after a few seconds as the ICE gathering will complete and no new candidates could be gather according to the standard.
**BUT**, although NOT standardized, Google WebRTC allows to configure a "GATHER_CONTINUALLY" (at least via native lib, I am not sure if it possible via browser), `GATHER_CONTINUALLY` makes the gathering to never finish, this can be very useful for mobile phones where it is relatively easy to start a call on mobile data and to connect to a WiFi network later on, or vice-verse. We are exploring ways to improve mobile user experience in Symphony and I am eager to enable this configuration on mobile.

### Future improvements
With `GATHER_CONTINUALLY` enabled, other non standardized mechanism Google implements is ICE renomination. Unlike `GATHER_CONTINUALLY`, which is a WebRTC peer connection configuration that seems to be accessible only through the native WebRTC, ICE renomination seems to be possible to be negotiated via SDP. There is a draft for the renomination https://datatracker.ietf.org/doc/html/draft-thatcher-ice-renomination-00 and it seems to be very useful together with `GATHER_CONTINUALLY`.

Other Google STUN attribute to explore is `STUN_ATTR_GOOG_NETWORK_INFO` this give us information about the network and it seems to be possible to infer if candidate is from 2G/3G/4G/5G, WIFI, ETHERNET and even if is using a VPN. Together with `CONTINUALLY_GATHERING` + ICE renomination, this attribute can be used to quick change to a better candidate once it is available

### DIFF
--- a/transport/TransportImpl.cpp
+++ b/transport/TransportImpl.cpp
@@ -2436,9 +2436,7 @@ void TransportImpl::onIceCandidateAccepted(ice::IceSession* session,
 
     for (auto& udpEndpoint : _rtpEndpoints)
     {
-        // static upper cast is a safer option to compare raw points as an
-        // offset might be added in case of multiple inheritance
-        // and ice::IceEndpoint* is not the first (although it not currently the case).
+        // static cast for type safe raw pointer comparison
         if (endpoint == static_cast<ice::IceEndpoint*>(udpEndpoint.get()))
         {
             udpEndpoint->registerListener(remoteCandidate.address, this);
@@ -2479,9 +2477,7 @@ void TransportImpl::onIceDiscardCandidate(ice::IceSession* session,
 
     for (auto& udpEndpoint : _rtpEndpoints)
     {
-        // static upper cast is a safer option to compare raw points as an
-        // offset might be added in case of multiple inheritance
-        // and ice::IceEndpoint* is not the first (although it not currently the case).
+        // static cast for type safe raw pointer comparison
         if (endpoint == static_cast<ice::IceEndpoint*>(udpEndpoint.get()))
         {
             udpEndpoint->unregisterListener(sourcePort, this);
--- a/transport/TransportImpl.cpp
+++ b/transport/TransportImpl.cpp
@@ -1076,22 +1076,10 @@ void TransportImpl::internalIceReceived(Endpoint& endpoint,
     auto dataReceiver = _dataReceiver.load();
     if (_rtpIceSession->isAttached(&endpoint))
     {
-        if (ice::isResponse(packet->get()) && _rtpIceSession->isResponseAuthentic(packet->get(), packet->getLength()))
-        {
-        }
-        else if (ice::isRequest(packet->get()) &&
-            _rtpIceSession->isRequestAuthentic(packet->get(), packet->getLength()))
-        {
-        }
-        else
+        if (!_rtpIceSession->isIceAuthentic(packet->get(), packet->getLength()))
         {
             // TODO attack, add metric for this
             return;
-        }
-
-        if (_rtpIceSession->canAcceptNewRemoteCandidate(timestamp, source))
-        {
-            endpoint.registerListener(source, this);
         }
 
         auto timeout = _rtpIceSession->nextTimeout(timestamp);
@@ -1857,6 +1845,7 @@ bool TransportImpl::setRemotePeer(const SocketAddress& target)
     {
         endpoint->registerListener(_peerRtpPort, this);
     }
+
     for (auto& endpoint : _rtcpEndpoints)
     {
         endpoint->registerListener(_peerRtcpPort, this);
@@ -1922,7 +1911,7 @@ void TransportImpl::doSetRemoteIce(const memory::AudioPacket& credentialPacket,
                     endpoint->registerDefaultListener(this);
                     _rtpEndpoints.push_back(endpoint);
 
-                    _rtpIceSession->addRemoteCandidate(candidate, endpoint.get());
+                    _rtpIceSession->addRemoteTcpPassiveCandidate(candidate, endpoint.get());
                 }
             }
         }
@@ -2434,13 +2423,66 @@ void TransportImpl::setTag(const char* tag)
     utils::strncpy(_tag, tag, sizeof(_tag));
 }
 
+void TransportImpl::onIceCandidateAccepted(ice::IceSession* session,
+    ice::IceEndpoint* endpoint,
+    const ice::IceCandidate& remoteCandidate)
+{
+    logger::info("candidate accepted: %s %s, %s - %s",
+        _loggableId.c_str(),
+        endpoint->getLocalPort().getFamilyString().c_str(),
+        ice::toString(endpoint->getTransportType()).c_str(),
+        endpoint->getLocalPort().toString().c_str(),
+        remoteCandidate.address.toString().c_str());
+
+    for (auto& udpEndpoint : _rtpEndpoints)
+    {
+        // static upper cast is a safer option to compare raw points as an
+        // offset might be added in case of multiple inheritance
+        // and ice::IceEndpoint* is not the first (although it not currently the case).
+        if (endpoint == static_cast<ice::IceEndpoint*>(udpEndpoint.get()))
+        {
+            udpEndpoint->registerListener(remoteCandidate.address, this);
+            return;
+        }
+    }
+
+    for (auto& udpEndpoint : _rtpEndpoints)
+    {
+        if (endpoint == static_cast<ice::IceEndpoint*>(udpEndpoint.get()))
+        {
+            udpEndpoint->registerListener(remoteCandidate.address, this);
+            return;
+        }
+    }
+
+    logger::error("Not possible to register listener as the endpoint was not found: %s %s, %s - %s",
+        _loggableId.c_str(),
+        endpoint->getLocalPort().getFamilyString().c_str(),
+        ice::toString(endpoint->getTransportType()).c_str(),
+        endpoint->getLocalPort().toString().c_str(),
+        remoteCandidate.address.toString().c_str());
+}
+
 void TransportImpl::onIceDiscardCandidate(ice::IceSession* session,
     ice::IceEndpoint* endpoint,
     const transport::SocketAddress& sourcePort)
 {
+    if (session->getState() <= ice::IceSession::State::CONNECTED)
+    {
+        logger::info("candidate discarded: %s %s, %s - %s",
+            _loggableId.c_str(),
+            endpoint->getLocalPort().getFamilyString().c_str(),
+            ice::toString(endpoint->getTransportType()).c_str(),
+            endpoint->getLocalPort().toString().c_str(),
+            sourcePort.toString().c_str());
+    }
+
     for (auto& udpEndpoint : _rtpEndpoints)
     {
-        if (endpoint == udpEndpoint.get())
+        // static upper cast is a safer option to compare raw points as an
+        // offset might be added in case of multiple inheritance
+        // and ice::IceEndpoint* is not the first (although it not currently the case).
+        if (endpoint == static_cast<ice::IceEndpoint*>(udpEndpoint.get()))
         {
             udpEndpoint->unregisterListener(sourcePort, this);
             return;
@@ -2449,7 +2491,7 @@ void TransportImpl::onIceDiscardCandidate(ice::IceSession* session,
 
     for (auto& udpEndpoint : _rtcpEndpoints)
     {
-        if (endpoint == udpEndpoint.get())
+        if (endpoint == static_cast<ice::IceEndpoint*>(udpEndpoint.get()))
         {
             udpEndpoint->unregisterListener(sourcePort, this);
             return;
--- a/transport/TransportImpl.cpp
+++ b/transport/TransportImpl.cpp
@@ -1968,8 +1968,8 @@ void TransportImpl::onIceCandidateChanged(ice::IceSession* session,
             _loggableId.c_str(),
             endpoint->getLocalPort().getFamilyString().c_str(),
             ice::toString(endpoint->getTransportType()).c_str(),
-            endpoint->getLocalPort().toString().c_str(),
-            sourcePort.toString().c_str());
+            endpoint->getLocalPort().toFixedString().c_str(),
+            maybeMasked(sourcePort).c_str());
 
         if (_srtpClient->getState() == SrtpClient::State::READY)
         {
@@ -1990,8 +1990,8 @@ void TransportImpl::onIceCandidateChanged(ice::IceSession* session,
             _loggableId.c_str(),
             endpoint->getLocalPort().getFamilyString().c_str(),
             ice::toString(endpoint->getTransportType()).c_str(),
-            endpoint->getLocalPort().toString().c_str(),
-            sourcePort.toString().c_str());
+            endpoint->getLocalPort().toFixedString().c_str(),
+            maybeMasked(sourcePort).c_str());
     }
 }
 
@@ -2019,8 +2019,8 @@ void TransportImpl::onIceStateChanged(ice::IceSession* session, const ice::IceSe
 
         logger::info("Pair selected: %s - %s  rtt:%" PRIu64 "us",
             _loggableId.c_str(),
-            candidatePair.first.address.toString().c_str(),
-            candidatePair.second.address.toString().c_str(),
+            candidatePair.first.address.toFixedString().c_str(),
+            maybeMasked(candidatePair.second.address).c_str(),
             rtt / utils::Time::us);
 
         if (_selectedRtp)
@@ -2431,8 +2431,8 @@ void TransportImpl::onIceCandidateAccepted(ice::IceSession* session,
         _loggableId.c_str(),
         endpoint->getLocalPort().getFamilyString().c_str(),
         ice::toString(endpoint->getTransportType()).c_str(),
-        endpoint->getLocalPort().toString().c_str(),
-        remoteCandidate.address.toString().c_str());
+        endpoint->getLocalPort().toFixedString().c_str(),
+        maybeMasked(remoteCandidate.address).c_str());
 
     for (auto& udpEndpoint : _rtpEndpoints)
     {
@@ -2471,8 +2471,8 @@ void TransportImpl::onIceDiscardCandidate(ice::IceSession* session,
             _loggableId.c_str(),
             endpoint->getLocalPort().getFamilyString().c_str(),
             ice::toString(endpoint->getTransportType()).c_str(),
-            endpoint->getLocalPort().toString().c_str(),
-            sourcePort.toString().c_str());
+            endpoint->getLocalPort().toFixedString().c_str(),
+            maybeMasked(sourcePort).c_str());
     }
 
     for (auto& udpEndpoint : _rtpEndpoints)
--- a/transport/ice/IceSession.cpp
+++ b/transport/ice/IceSession.cpp
@@ -1524,14 +1524,17 @@ int64_t IceSession::processTimeout(const uint64_t now)
     }
 
     DBGCHECK_SINGLETHREADED(_mutexGuard);
-    bool hasFailedCandidates = false;
+    bool hadInProgressBeforeProcess = false;
+    bool hasInProgressAfterProcess = false;
     for (auto it = _candidatePairs.rbegin(); it != _candidatePairs.rend(); ++it)
     {
         auto* candidatePair = it->get();
 
         const auto currentState = candidatePair->state;
+        hadInProgressBeforeProcess = hadInProgressBeforeProcess || currentState == ProbeState::InProgress;
         candidatePair->processTimeout(now);
-        hasFailedCandidates = hasFailedCandidates || candidatePair->state == ProbeState::Failed;
+        hasInProgressAfterProcess = hasInProgressAfterProcess || candidatePair->state == ProbeState::InProgress;
+
         if (currentState == ProbeState::InProgress && candidatePair->state == ProbeState::Failed)
         {
             if (candidatePair->localCandidate.transportType == ice::TransportType::TCP)
@@ -1564,6 +1567,12 @@ int64_t IceSession::processTimeout(const uint64_t now)
                 }
             }
         }
+    }
+
+    const bool lastInProgressHasCompleted = hadInProgressBeforeProcess && !hasInProgressAfterProcess;
+    if (lastInProgressHasCompleted && _state == State::CONNECTED)
+    {
+        removeUnviableRemoteCandidates(now);
     }
 
     stateCheck(now);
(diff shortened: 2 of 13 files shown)


transport/TransportImpl.cpp
```diff
@@ -1076,24 +1076,12 @@ void TransportImpl::internalIceReceived(Endpoint& endpoint,
...
         {
             // TODO attack, add metric for this
             return;
         }
 
-        if (_rtpIceSession->canAcceptNewRemoteCandidate(timestamp, source))
-        {
-            endpoint.registerListener(source, this);
-        }
-
```
RicardoMDomingues, line 1096: We accept can accept candidates any time. The IceSession is going to call "onIceAccpted" and it will be registered on that moment

transport/ice/IceSession.h
```diff
@@ -292,28 +307,25 @@ class IceSession
...
         uint64_t _transmitInterval;
         int _replies;
         IceError _errorCode;
         uint64_t _minRtt;
-        const std::string& _name;
         std::deque<StunTransaction> _transactions; // TODO replace with inplace circular container
-        StunTransactionIdGenerator& _idGenerator;
-        const IceConfig& _config;
-        const SessionCredentials& _credentials;
-        crypto::HMAC& _hmacComputer;
```
RicardoMDomingues, line 304: Reduce IceCandidatePair memory footprint. We store only a pointer to IceSession so we can have access fom everything we need

transport/ice/IceSession.cpp
```diff
@@ -472,11 +493,252 @@ bool IceSession::isResponseAuthentic(const void* data, size_t len) const
...
+    if (!(_state == State::CONNECTING || _state == State::CONNECTED))
+    {
+        return false;
+    }
+
+    auto* pair = findCandidatePair(localEndpoint, remotePort);
+    if (pair)
+    {
+        // Only send response after accept! Otherwise we could drop early media
+        pair->accept();
```
RicardoMDomingues, line 566: first accept pair. This will call the listener to register the endpoint (when it is not "accepted" already). Only then send response to be sure clients considers this a valid pair when we have already register the peer on udp endpoint

transport/ice/IceSession.cpp
```diff
@@ -472,11 +493,252 @@ bool IceSession::isResponseAuthentic(const void* data, size_t len) const
...
+        // I don't see why this should happen but if we are receiving a probe from it, perhaps there is no reason to be
+        // discarded. Let's add it again and log a warn
+        const auto* localEndpointInfo = findEndpointInfo(localEndpoint);
+        assert(localEndpointInfo);
+        if (!localEndpointInfo)
+        {
+            // We can't accept endpoint that are not attached. This should be an error as it is not supposed to
+            // happen
+            // This perhaps should send an error instead of just ignore
+            return false;
```
RicardoMDomingues, line 585: Perhaps this should send an error response. It is not supposed to receive requests from a local endpoint that is not attached

transport/ice/IceSession.cpp
```diff
@@ -472,11 +493,252 @@ bool IceSession::isResponseAuthentic(const void* data, size_t len) const
...
+
+    // If we are here, it is a new candidate
+
+    if (_state == State::CONNECTED || _remoteCandidates.size() >= _config.maxCandidateCount)
+    {
+        // When we see a new candidate and the ICE is CONNECTED, this can be an early sign that network state
+        // has changed on client. Therefore, we will try to remove some candidates that has not being received
+        // probes nor media for a while to ensure we can had new candidates. The removing will be more
+        // aggressive if the _remoteCandidates is full
+        removeNonReadableRemoteCandidates(now);
```
RicardoMDomingues, line 611: If the candidate list is full or the state is connected (perhaps is an early indication that something has changed on client with a new candidate being available). We are going to try to reduce the candidate list

transport/ice/IceSession.cpp
```diff
@@ -472,11 +493,252 @@ bool IceSession::isResponseAuthentic(const void* data, size_t len) const
...
+                    _logId.c_str(),
+                    useCandidatePair->localCandidate.address.toString().c_str(),
+                    useCandidatePair->remoteCandidate.address.toString().c_str());
+                if (useCandidatePair->state == ProbeState::Succeeded)
+                {
+                    if (_nomination)
+                    {
+                        // Aggressive nomination
+                        // Then we will clean the old nominated flag
+                        _nomination->nominated = false;
```
RicardoMDomingues, line 708: Add support to agressive nomination when we are CONTROLLED. The code before it was already accepting new nominations but was not clearing the flag

transport/ice/IceSession.cpp
```diff
@@ -472,11 +493,252 @@ bool IceSession::isResponseAuthentic(const void* data, size_t len) const
...
     const transport::SocketAddress& remotePort,
     const StunMessage& msg,
     const uint64_t now)
 {
+    if (_state >= State::FAILED)
+    {
+        // SMB does not have a way to restart ICE. Then we should not respond more to probes after we have a failed
+        // state otherwise peer might think the ICE connection is still good and then does not try any recover
+        // mechanism. So if it is FAILED, peer MUST fail too
+        return;
```
RicardoMDomingues, line 739: As SMB does not have mechanism to report its ICE state, it can't respond to stuns when it is in FAIL state, otherwise clients might not fail and then they don't trigger the recover

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
...
-                        candidatePair->remoteCandidate.address);
-                }
-                removeRemoteCandidate(candidatePair->remoteCandidate);
-                removeCandidatePair(candidatePair);
-            }
-            else if (candidatePair->remoteCandidate.type == IceCandidate::Type::PRFLX)
-            {
-                // port may be recycled in firewall anyway
                 candidatePair->state = ProbeState::Failed;
-                removeRemoteCandidate(candidatePair->remoteCandidate);
```
RicardoMDomingues, line 1115: We no longer delete PRFLX candidates, they can be our backup candidates later on. WebRTC will continue probe backup candidates each 25 seconds, prevent port to be recycled in firewall

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
...
+                onCandidatePairRemoved(candidatePair);
+                std::iter_swap(it, --itEnd);
+                continue;
             }
         }
+        // If didn't fail, then it can be a backup in case the nominated becomes unreachable
+        else if (candidatePair->remoteCandidate.type == IceCandidate::Type::PRFLX &&
```
RicardoMDomingues, line 1299: We do remove PRFLX candidates if they FAILED. Probably we could delete SRFLX and HOST too. If they fail, there is low chance to them to start be used later on

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
...
+        // Let's keep the nominated regardless the current state
+        if (candidatePair->nominated)
+        {
+            addCandidateToListIfNotPresent(_remoteCandidates, candidatePair->remoteCandidate);
+            ++it;
+            continue;
+        }
+
+        // Keep HOST and SRFLX if we are not in aggressive removing.
+        if (!aggressiveRemove && candidatePair->remoteCandidate.type != IceCandidate::Type::PRFLX)
```
RicardoMDomingues, line 1344: If the candidates list is not full `aggressiveRemove = false`, we will keep SRFLX and HOST candidates

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
...
+        {
+            addCandidateToListIfNotPresent(_remoteCandidates, candidatePair->remoteCandidate);
+            ++it;
+            continue;
+        }
+
+        if (candidatePair->isReadable(now))
```
RicardoMDomingues, line 1351: We will keep the candidate if we receive something from peer in latest 45 seconds. WebRTC clients probes backup candidates each 25 seconds. Then after 45 seconds chances are low that is still a working pair

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
...
+        candidatePair->state = ProbeState::Failed;
+        std::iter_swap(it, --itEnd);
+        onCandidatePairRemoved(candidatePair);
+    }
+
+    _candidatePairs.erase(itEnd, _candidatePairs.end());
+
+    // If after the first scan the list it still bigger or equal to _config.maxCandidateCount,
+    // then we will be even more aggressive on removing candidates
+    if (_remoteCandidates.size() <= _config.maxCandidateCount)
```
RicardoMDomingues, line 1374: If after first round, we still have a full candidate list, then we going to try to release some of them. I am not truly confident that we should do this but I am not sure if rejecting a new candidate is actually a better option.

Then the algorithm fist sort the candidatesPair. First the dominated, then the ones in a readable state ordered by priority (where we receive packet in the latest 45 seconds. Then the other ones ordered by priority.

Then we going to read them to `_remoteCandidates ` until a point that the list is not full and remove the lower priority ones that didn't fit

transport/ice/IceSession.cpp
```diff
@@ -1613,20 +1965,22 @@ void IceSession::CandidatePair::processTimeout(const uint64_t now)
...
     {
         if (nominated &&
-            (utils::Time::diffGE(receptionTimestamp, now, _config.hostProbeTimeout * utils::Time::ms) ||
+            (utils::Time::diffGE(receptionTimestamp, now, _iceSession->_config.hostProbeTimeout * utils::Time::ms) ||
                 pendingTransactionCount() > 2))
         {
-            logger::debug("restart prob due to lack of inbound traffic", _name.c_str());
-            restartProbe(now);
+            logger::info("restart prob due to lack of inbound traffic", _iceSession->_logId.c_str());
+            restartProbe(now, false);
```
RicardoMDomingues, line 1972: When we see lack of traffic on the nominated, we will restart probe without remove the nomination. only after a few time, we consider unreachable and then we will remove the nomination and change the state to `CONNECTING`

transport/ice/IceSession.cpp
```diff
@@ -1185,20 +1505,15 @@ int64_t IceSession::processTimeout(const uint64_t now)
...
-                        candidatePair->localEndpoint.endpoint,
-                        candidatePair->remoteCandidate.address);
-                }
                 removeCandidatePair(candidatePair);
             }
         }
-        else if (_nomination == candidatePair && candidatePair->state != ProbeState::Succeeded)
+        else if (_nomination == candidatePair && candidatePair->nominatedInProgressNextTimedout(now) == 0)
         {
-            logger::debug("found nominated pair in state %s", _logId.c_str(), toString(candidatePair->state));
```
RicardoMDomingues, line 1199: After timeout of the nominated candidate be in InProgress, we then remove the nomination and change the state to CONNECTIN

transport/TransportImpl.cpp
```diff
@@ -2434,13 +2423,66 @@ void TransportImpl::setTag(const char* tag)
     utils::strncpy(_tag, tag, sizeof(_tag));
 }
 
+void TransportImpl::onIceCandidateAccepted(ice::IceSession* session,
+    ice::IceEndpoint* endpoint,
+    const ice::IceCandidate& remoteCandidate)
+{
+    logger::info("candidate accepted: %s %s, %s - %s",
```
olofkallander, line 2430: Maybe it is not consistent through code, but I think there was some reasoning that PII info like IP should only be logged at debug level. So it is generally not available in the cloud logs.

transport/TransportImpl.cpp
```diff
@@ -2434,13 +2423,66 @@ void TransportImpl::setTag(const char* tag)
...
+            endpoint->getLocalPort().getFamilyString().c_str(),
+            ice::toString(endpoint->getTransportType()).c_str(),
+            endpoint->getLocalPort().toString().c_str(),
+            sourcePort.toString().c_str());
+    }
+
     for (auto& udpEndpoint : _rtpEndpoints)
     {
-        if (endpoint == udpEndpoint.get())
+        // static upper cast is a safer option to compare raw points as an
```
olofkallander, line 2482:
```suggestion
        // static cast for type safe raw pointer comparison
```

transport/ice/IceSession.cpp
olofkallander: yes perhaps. It is easier to debug and remote side understands the candidate is undesired. A sendResponse with error should suffice.

transport/ice/IceSession.cpp
```diff
@@ -472,11 +493,252 @@ bool IceSession::isResponseAuthentic(const void* data, size_t len) const
...
+        // Only send response after accept! Otherwise we could drop early media
+        pair->accept();
+        sendResponse(localEndpoint, remotePort, 0, msg, now);
+        return true;
+    }
+
+    const auto* remoteCandidate = findCandidateWithUdpAddress(_remoteCandidates, remotePort);
+    if (remoteCandidate)
+    {
+        // if it is not a new candidate but we don't have a candidate that could mean we are receiving
```
olofkallander, line 574: I wonder if it is better to avoid this logic. If we discarded the candidate, it should be for a reason. Reviving it could be a problem in case we do want to ignore certain candidates. We should get the pruning logic right instead. Perhaps lopg this but just reject the candidate.

transport/ice/IceSession.h
```diff
@@ -292,28 +307,25 @@ class IceSession
         uint64_t pendingRequestAge(uint64_t now) const;
         size_t pendingTransactionCount() const;
 
+        IceSession* _iceSession;
```
olofkallander, line 310: Good change, but is there a case where this needs to be a pointer and not & ?

transport/TransportImpl.cpp
```diff
@@ -1076,24 +1076,12 @@ void TransportImpl::internalIceReceived(Endpoint& endpoint,
...
-        else if (ice::isRequest(packet->get()) &&
-            _rtpIceSession->isRequestAuthentic(packet->get(), packet->getLength()))
-        {
-        }
-        else
+        if (!_rtpIceSession->isIceAuthentic(packet->get(), packet->getLength()))
         {
             // TODO attack, add metric for this
             return;
         }
```
olofkallander, line 1083: Just trying to understand the situation the change is for. 
Before: new candidates were not accpted after reflexiveProbeTimeout (15s) had passed, unless the nominated probe did not receive responses for 5s, we did accept new candidates regardless of how long into the session we were.
What is the new situation this fix is for? is it to maintain multiple alternative candidates or otherwise the remote side will discard them too, and there is not fail over when nominated fails?

transport/ice/IceSession.h
```diff
@@ -239,31 +250,36 @@ class IceSession
...
         int64_t nextTimeout(uint64_t now) const;
         void processTimeout(uint64_t now);
 
-        void restartProbe(const uint64_t now);
+        void restartProbe(const uint64_t now, bool releaseNomination);
+        int64_t nominatedInProgressNextTimedout(const uint64_t now) const;
         void send(uint64_t now);
         uint64_t getPriority(IceRole role) const;
         bool isFinished() const { return state == ProbeState::Succeeded || state == ProbeState::Failed; }
+        bool isReadable(uint64_t now) const
```
olofkallander, line 269: A better name is perhaps "isViable".
Meaning: "Capable of success or continuing effectiveness"

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
...
+        }
+
+        addCandidateToListIfNotPresent(_remoteCandidates, candidatePair->remoteCandidate);
+        ++it;
+    }
+
+    _candidatePairs.erase(itEnd, _candidatePairs.end());
+}
+
+void IceSession::removeNonReadableRemoteCandidates(uint64_t now)
```
olofkallander, line 1321: removeUnviableRemoteCandidates

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
 
 void IceSession::freezePendingProbes(uint64_t now)
 {
-    for (auto it = _candidatePairs.rbegin(); it != _candidatePairs.rend(); ++it)
+    // We are going to clear all remote candidates without candidatePairs
+    _remoteCandidates.clear();
```
olofkallander, line 1274: There is a bit unsafe if it happens before all candidate pairs have been created because probes are added at pace. Remote candidates towards end of list will be lost.

transport/ice/IceSession.cpp
```diff
@@ -1091,31 +1270,172 @@ void IceSession::stateCheck(const uint64_t now)
...
+    _candidatePairs.erase(itEnd, _candidatePairs.end());
+}
+
+void IceSession::removeNonReadableRemoteCandidates(uint64_t now)
+{
+    // If we are in the limit, we will be more aggressive on removing candidates
+    // by removing HOST and SRFLX candidates that didn't work or are inactive for long period.
+    // Wen the list is not full, we will only remove PRFLX
+    const bool aggressiveRemove = _remoteCandidates.size() >= _config.maxCandidateCount;
+    _remoteCandidates.clear();
```
olofkallander, line 1327: Like I commented above, I think it is better to keep all and let them be removed as the probes fail.

transport/ice/IceSession.h
RicardoMDomingues: Yes, I can change to a reference

transport/TransportImpl.cpp
RicardoMDomingues: Is the description I added helped you to understand the change? The problem was mainly caused by removing PRFLX candidates and only accept them when we were in CONNECTING or until to reflexiveProbeTimeout. The problem is SMB usually takes longer to realise the nominated is not reachable than WebRTC client. And WebRTC client will probe immediately all candidates when it detects the dominated is not reachable. At that moment, SMB can be on CONNECTED state still and not re-add the candidates. When it switch to CONNECTING, it will not have PRFLX backup candidates.

But, in first place, the problem was caused by removing the PRFLX candidates after nomination. Continue accepting is just an improvement for support GATHER_CONTINUALLY where clients can gather candidates anytime and we want to add them as backup.

transport/ice/IceSession.cpp
RicardoMDomingues: Candidate pairs are created right after creating PRFX candidate. We are only removing PRFLX candidates, meaning that we are not deleting anything does not have probes (exception is for the TCP canidates that we don't care if they are PRFLX as we are agressive on prune tcp connection).

But yes, we could let the probes to complete and delete the failed ones, so we keep only the candidates that work. I will change that.

Regarding TCP, transport is already deleting all TCP endpoints then this logic of keeping TCP that does not have valid probes on UDP is not needed as transport will kill the socket.

RicardoMDomingues: I rename this method to not crate confusing. It is now `pruneCandidatesAfterConnected` (and it should be called once the states gets connected).

This removes:
 - Failed candidates (regardless if they are PRFLX, SRFLX, HOST, etc)
 - TCP candidates. TransportImpl will delete the endpoints anyway
 
 This freezes the Succeeded candidates if not nominated (this will make SMB stops to probe them). Its completely to the client to keep them 'viable' by continue probing them (I guess C9 is much more aggressive on freeing candidates so let the clients to decide) 
 
 It does not touch on InProgress candidate pairs. They must Fail or Succeed  at some point. When a candidate pair Succeed after state is CONNECTED, the state transits to frozen instead.
 
I changed the method `processTimeout` to call `removeUnviableRemoteCandidates` when the last InProgress candidate completes its state and ICE state is CONNECTED (so we can remove the unviable candidates in the end without run the method each time a candidate fails)

transport/ice/IceSession.cpp
RicardoMDomingues: I had simplifies method `removeUnviableRemoteCandidates`. We don't have that aggressive removing logic any longer. We delete every candidate pair that is no viable

The remote candidates is re-added to the list each time time we decide to keep a candidate pair that contains a candidate the is not in `_remoteCandidates`. So at the end, we will have only _remoteCandidates that have viable candidate pairs

transport/ice/IceSession.cpp
RicardoMDomingues: I will do that. But as we are always trying to keep `remoteCandidate` short, I guess this is unlikely (if not impossible) to detect. We will see like if it was a new PRFLX candidate

reddvl1980 (approved):

PR #379, opened 2024-09-11
MEDIA-3073: allow crash on SIGPIPE and stop log to stdout

reddvl1980:


RicardoMDomingues (approved):

main.cpp
```diff
@@ -18,17 +18,14 @@ std::unordered_map<int32_t, struct sigaction> oldSignalHandlers;
...
-    logger::errorImmediate("Fatal signal %d, %d stack frames.", "fatalSignalHandler", signalId, size);
-    logger::logStackImmediate(array, size, "fatalSignalHandler");
-    logger::error("Fatal signal %d, %d stack frames.", "fatalSignalHandler", signalId, size);
-
-    logger::flushLog();
-
-    if (signalId == SIGPIPE)
-    {
-        return;
+    if (signalId != SIGPIPE) {
```
RicardoMDomingues, line 21: Instead of doing this can we just disable the stdout logging on SIGPIPE before logging. So we still have the logs and  back traces on log file?

PR #384, opened 2024-10-13
split cmakelist

olofkallander:
Split cmakelist into test and main. Can this work for the build setup?


CMakeLists.txt
```diff
@@ -472,239 +472,14 @@ add_executable(smb
...
-
-add_test(AllUnitTests UnitTest)
-add_test(AllUnitTests UnitTest2)
-
-add_executable(LoadTest ${LOAD_TEST_FILES}
-    test/load_test_main.cpp)
-
-target_link_libraries(LoadTest testlib gtest gmock)
+add_subdirectory(test)
+set_target_properties(UnitTest PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR} ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR} RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
```
olofkallander, line 480: This is to make Test executables end up in build top folder. 
However, maybe we should consider allowing the build folder to be structured as cmake default:
build
  bin
  _deps
  libs

test/CMakeLists.txt
```diff
@@ -0,0 +1,234 @@
+include(FetchContent)
+
+# Download and unpack googletest at configure time
+configure_file(CMakeLists-gtest.txt.in googletest-download/CMakeLists.txt)
```
olofkallander, line 4: This should be done with fetchContent but it would require the cmake to be upgraded on all docker images first.

PR #386, opened 2024-10-28
ROC fix

olofkallander:
Libsrtp has to start RTP decryption on ROC=0. If the sender starts near roll over 0->1, and those first critical packets are lost, the media stream will never be possible to decrypt.
This PR removes that risk by synthesizing an RTP packet with ROC=0, if the first decryption attempt fails. It will then be able to try with ROC 1,2,3, which covers a window of several minutes. Thus, the decryption can always start, also when all ROC 0 packets are lost.
It may also happen in mid call if a new stream starts at the time of network problems.

Another bug fix was also made for data channel setup in case of packet loss. The COOKIE-ECHO retransmission had wrong length, and SCTP establishment failed.


transport/sctp/SctpAssociationImpl.cpp
```diff
@@ -184,7 +184,7 @@ size_t SctpAssociationImpl::GenericCookie::maxSize() const
 void SctpAssociationImpl::GenericCookie::set(const ChunkParameter& param)
 {
     std::memcpy(cookie, param.data(), std::min(static_cast<size_t>(param.dataSize()), sizeof(cookie)));
-    length = param.length.get();
+    length = param.dataSize();
```
olofkallander, line 187: bug fix for SCTP establishment

RicardoMDomingues: Nice, I have been some sporadic failures to open datachannels and I added logs around the Cookie negation. Perhaps this was the reason

RicardoMDomingues (approved):

axeleriksson147:
I think this PR or https://github.com/finos/SymphonyMediaBridge/pull/385 introduced some changes that broke our e2e-testing when media is restarted. The screenshotted image of screen share or video gets very funky. Haven't dug into why, but it points towards this. FYI.

@olofkallander

PR #397, opened 2025-01-09
RTC-15334: Fix crash caused by a race condition

RicardoMDomingues:
I believe this fix the crash we have seen on compute hmac.

The crash happens to a race condition. STUN packets that are addressed to ProbeServer are being processed on Endpoint jobQueue. If multiple STUN packets arrives in different endpoint (udp ipv4, udp ipv6, tcp ipv4, tcp ipv6) the race can happen on `_hmacComputer`.

To fix that, ProbeServer now has it's own jobQueue

### DIFF
--- a/transport/ProbeServer.cpp
+++ b/transport/ProbeServer.cpp
@@ -5,14 +5,18 @@
 #include "transport/ice/IceSession.h"
 #include "transport/ice/Stun.h"
 #include "utils/ContainerAlgorithms.h"
+#include "utils/Function.h"
 #include "utils/Time.h"
 #include <unistd.h>
 
 namespace transport
 {
-ProbeServer::ProbeServer(const ice::IceConfig& iceConfig, const config::Config& config)
+ProbeServer::ProbeServer(const ice::IceConfig& iceConfig,
+    const config::Config& config,
+    jobmanager::JobManager& jobmanager)
     : _iceConfig(iceConfig),
       _config(config),
+      _jobQueue(jobmanager, 1024),
       _queue(1024),
       _maintenanceThreadIsRunning(true),
       _maintenanceThread(new std::thread([this] { this->run(); }))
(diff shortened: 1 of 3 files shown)


transport/ProbeServer.cpp
```diff
@@ -274,11 +281,40 @@ void ProbeServer::replyStunOk(Endpoint& endpoint, const SocketAddress& destinati
         response.addFingerprint();
 
         endpoint.sendStunTo(destination, response.header.transactionId.get(), &response, response.size(), timestamp);
+        return true;
     }
-    else
+
+    if (endpoint.getTransportType() != ice::TransportType::UDP)
```
RicardoMDomingues, line 287: I believe this is another bug. If we received something that was not properly authenticated, we would close UdpEndpoint which is shared port 10000

reddvl1980 (approved):
LGTM

olofkallander (commented):
Nice catch. I guess this contention on hmac computer becomes more severe with use of ipv6 and higher traffic. 
Most of the time a client will not probe at all before call and if it does, it fires at udp ipv4 before it does tcp v4, if at all. Even if multiple clients probe simultaneously they are serialized over the udp endpoint recvjobqueue. But with ipv6 udp there are two parallell paths.

olofkallander (approved):

RicardoMDomingues:
> Nice catch. I guess this contention on hmac computer becomes more severe with use of ipv6 and higher traffic. Most of the time a client will not probe at all before call and if it does, it fires at udp ipv4 before it does tcp v4, if at all. Even if multiple clients probe simultaneously they are serialized over the udp endpoint recvjobqueue. But with ipv6 udp there are two parallell paths.

When I was doing this I was thinking in add two (or more) queues with their own hmacComputer instance and use round robin too avoid contention. Decided that probably is not worth, as you said, this is something that clients do once in a while, then there will be not much pressure on this endpoint I believe.  And everything gets conteined already by endpoint queues, so don't think this will cause much delays (only when process different endpoints in simultaneous), still the client should use the lowest response time after an X number of stun responses, which hopefully is enough to avoid wrong decision due contention

PR #398, opened 2025-01-11
H264inclient

olofkallander:
Added h264 support in simple client and allow selection of h264 at allocation of conference if h264 is also enabled in config.json. 
Default is vp8.

### DIFF
--- a/api/AllocateConference.h
+++ b/api/AllocateConference.h
@@ -10,6 +10,7 @@ struct AllocateConference
 {
     utils::Optional<uint32_t> lastN;
     bool useGlobalPort = true;
+    bool useH264 = false;
 };
 
 } // namespace api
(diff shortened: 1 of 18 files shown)


api/AllocateConference.h
```diff
@@ -10,6 +10,7 @@ struct AllocateConference
 {
     utils::Optional<uint32_t> lastN;
     bool useGlobalPort = true;
+    bool useH264 = false;
```
RicardoMDomingues, line 13: can we do it a little more generic. So if we want to support new video codec ion future (like vp9 or vp10). We will flood this of booleans.

I believe we should have a filed more generic, like:
`videoCodec = "default | vp8 | h264" // default is the one that is on smb config`

If given an unsupported value, http request should fail with 400.

This way we could add support to new codecs without changing API

PR #387, opened 2024-11-02
Dockerupgrade

olofkallander:
This is an update of Dockerfiles to use 
openssl 3.4.0
cmake
libsrtp 2.6.0
building and pushing new docker images is needed for this to have any effect on the builds.
Building a much older version of smb may not be possible on the new docker images.

PR #399, opened 2025-01-14
Dockerupgrade (#387)

RicardoMDomingues:
* aws-docker update

* el8 docker update

* ubuntu-focal

* jammy

* focal quicker cmake

* cp ssl 1 also

### DIFF
--- a/.github/workflows/pr-unit-test.yml
+++ b/.github/workflows/pr-unit-test.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         include:
           - os: el8
-            image: d9f98e8
+            image: 2dd8104
             build:
               type: Release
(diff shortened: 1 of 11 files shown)


.github/workflows/pr-unit-test.yml
```diff
@@ -23,31 +23,31 @@ jobs:
...
             build:
               type: TCheck
           - os: el8
-            image: latest
+            image: d9f98e8
             build:
               type: DCheck
           - os: el8
-            image: latest
+            image: d9f98e8
```
RicardoMDomingues, line 50: Change this to specific images so when we build new image does not affect the current master and if we update some library that requires code changes, does not break previous commit builds

olofkallander (approved):
LGTM

PR #401, opened 2025-02-11
MEDIA-3445: Reduce memory consumption for audio only conferences

RicardoMDomingues:


concurrency/MpmcQueue.h
```diff
@@ -249,13 +282,21 @@ class MpmcQueue
...
     // Layout below must be maintained for performance. The non atomic member variables may not be place before read
     // cursor.
-    std::atomic<VersionedIndex> _readCursor;
-    uint32_t _cacheLineSeparator1[14];
-    std::atomic<VersionedIndex> _writeCursor;
-    uint32_t _cacheLineSeparator2[14];
-    const uint32_t _entriesPerCacheline;
-    const uint32_t _capacity;
-    const size_t _blockSize;
+    alignas(64) std::atomic<VersionedIndex> _readCursor;
```
RicardoMDomingues, line 285: Replace the explicit line sepatator fields and forced the alignment on 64 bits. This is a bit more resilient as before we couldn't have sure if `readCursor` was not sharing the cache line with another atomic members in a different object. `alignas(64)` force it to be the first address on the cache line

concurrency/MpmcQueue.h
```diff
@@ -249,13 +282,21 @@ class MpmcQueue
...
+
+    // Special Cell state for when queue has zero size
+    // Null Entry is not written and it is only read it when queue is zero size. So it can live in same cache line of
+    // _readCursor without performance impact
+    const EntrySate _nullEntry = CellState::State::nullSlot;
```
RicardoMDomingues, line 290: This is in the same cache line of _readCursor to save a little of memory. This value is unchangeable   and it only read when _readCursor is also unchangeable (ie, when the capacity is zero)

concurrency/MpmcQueue.h
```diff
@@ -249,13 +282,21 @@ class MpmcQueue
...
+
+    alignas(64) std::atomic<VersionedIndex> _writeCursor;
+
+    // Although they are const. Cache can be invalidated for the reader when _writeCursor
+    // Perhaps is not a big issue (and even compile can optimize it this const values)
+    // but forcing a new cache line also ensures tha nothing is placed in this cache line after
+    // MpmcQueue boundary (which we can predict if can affect performance
+    // of _writeCursor)
+    alignas(64) const uint32_t _capacity;
     Entry* const _elements;
```
RicardoMDomingues, line 300: I am not convinced that separating this 2 fields from `_writeCursor` has a significant impact on the performance as they are read only fields (so they don't cause cache trashing), maybe they can suffer from cache trashing when they are being red but as they are const filed, I think compiler has way to optimize it (or even we could to store them on the stack memory when on the begining of each method that needs them)

concurrency/MpmcQueue.h
```diff
@@ -69,21 +101,19 @@ class MpmcQueue
 public:
     typedef T value_type;
     explicit MpmcQueue(uint32_t capacity)
-        : _entriesPerCacheline((63 + sizeof(Entry)) / sizeof(Entry)),
-          _capacity(_entriesPerCacheline * ((capacity + _entriesPerCacheline - 1) / _entriesPerCacheline)),
-          _blockSize(memory::page::alignedSpace(_capacity * sizeof(Entry))),
-          _elements(reinterpret_cast<Entry*>(memory::page::allocate(_blockSize)))
+        : _capacity(
```
olofkallander, line 104: Isn't it better to have an assert for capacity==0? What is the point of zero length queues?

concurrency/MpmcQueue.h
```diff
@@ -19,19 +19,49 @@ class MpmcQueue
...
             emptySlot = 0,
-            committed
+            committed,
+            nullSlot // this a especial slot for when queue is zero capacity, Which make the slot unreadable and
+                     // unwritable
         } state = emptySlot;
     };
 
-    struct Entry
+    struct EntrySate
```
olofkallander, line 28: EntryState

concurrency/MpmcHashmap.h
```diff
@@ -180,10 +181,10 @@ class MpmcHashmap32
 
     explicit MpmcHashmap32(size_t maxElements) : _end(0), _capacity(maxElements), _index(maxElements * 4)
     {
-        void* mem = memory::page::allocate(memory::page::alignedSpace(_capacity * sizeof(Entry)));
+        void* mem = _capacity ? memory::page::allocate(memory::page::alignedSpace(_capacity * sizeof(Entry))) : nullptr;
```
olofkallander, line 184: would it not be easier with
if (_capacity == 0){ _elements = nullptr; return;}

concurrency/MpmcHashmap.h
```diff
@@ -90,6 +90,7 @@ class MpmcHashmap32
         tombstone
     };
 
+public:
```
olofkallander, line 93: maybe remove the next public:

bridge/Mixer.h
```diff
@@ -94,12 +94,14 @@ class Mixer
...
         const std::vector<api::SsrcPair>& videoPinSsrcs,
-        bool useGlobalPort);
+        bool useGlobalPort,
+        bool disableVideo);
 
     virtual ~Mixer() = default;
 
     void markForDeletion();
     bool isMarkedForDeletion() const { return _markedForDeletion; }
+    bool hasVideoDisabled() const { return _videoDisabled; }
```
olofkallander, line 104: isVideoDisabled() const {return _videoSsrcs.empty();}

bridge/Mixer.h
```diff
@@ -94,12 +94,14 @@ class Mixer
         const std::vector<uint32_t>& audioSsrcs,
         const std::vector<api::SimulcastGroup>& videoSsrcs,
         const std::vector<api::SsrcPair>& videoPinSsrcs,
-        bool useGlobalPort);
+        bool useGlobalPort,
+        bool disableVideo);
```
olofkallander, line 98: not really necessary as videoSsrcs will be empty vector and then you cannot do video.

bridge/engine/EngineMixer.cpp
```diff
@@ -31,24 +31,25 @@ EngineMixer::EngineMixer(const std::string& id,
...
       _loggableId("EngineMixer"),
       _jobManager(jobManager),
       _engineSyncContext(engineSyncContext),
       _messageListener(messageListener),
       _incomingBarbellSctp(128),
       _incomingForwarderAudioRtp(maxPendingPackets),
-      _incomingRtcp(maxPendingRtcpPackets),
-      _incomingForwarderVideoRtp(maxPendingPackets),
+      _incomingRtcp(isVideoDisabled ? maxPendingRtcpPacketsVideoDisabled : maxPendingRtcpPackets),
+      _incomingForwarderVideoRtp(isVideoDisabled ? 0 : maxPendingPackets),
```
olofkallander, line 44: videoSsrcs.empty() ?

concurrency/MpmcQueue.h
```diff
@@ -249,13 +282,21 @@ class MpmcQueue
 
     // Layout below must be maintained for performance. The non atomic member variables may not be place before read
```
olofkallander, line 283: placed

concurrency/MpmcQueue.h
```diff
@@ -249,13 +282,21 @@ class MpmcQueue
...
+    alignas(64) std::atomic<VersionedIndex> _readCursor;
+
+    // Special Cell state for when queue has zero size
+    // Null Entry is not written and it is only read it when queue is zero size. So it can live in same cache line of
+    // _readCursor without performance impact
+    const EntrySate _nullEntry = CellState::State::nullSlot;
+
+    alignas(64) std::atomic<VersionedIndex> _writeCursor;
+
+    // Although they are const. Cache can be invalidated for the reader when _writeCursor
```
olofkallander, line 294: // place _capacity in next cacheline to avoid interference with writeCursor on size() and capacity()

concurrency/MpmcQueue.h
```diff
@@ -19,19 +19,49 @@ class MpmcQueue
...
 
         T& value() { return reinterpret_cast<T&>(data); }
-        std::atomic<CellState> state;
-        uint8_t data[sizeof(T)];
+        alignas(alignof(T)) uint8_t data[sizeof(T)];
     };
 
+    Entry* nullEntry()
+    {
+        // Make _elements to point to nullEntry (which is both unreadable and unwritable)
```
olofkallander, line 47: // Dummy entry for empty queues.

concurrency/MpmcQueue.h
```diff
@@ -19,19 +19,49 @@ class MpmcQueue
         enum State : uint32_t
         {
             emptySlot = 0,
-            committed
+            committed,
+            nullSlot // this a especial slot for when queue is zero capacity, Which make the slot unreadable and
```
olofkallander, line 23: // special state only for zero size queue

concurrency/MpmcQueue.h
```diff
@@ -249,13 +282,21 @@ class MpmcQueue
...
-    std::atomic<VersionedIndex> _readCursor;
-    uint32_t _cacheLineSeparator1[14];
-    std::atomic<VersionedIndex> _writeCursor;
-    uint32_t _cacheLineSeparator2[14];
-    const uint32_t _entriesPerCacheline;
-    const uint32_t _capacity;
-    const size_t _blockSize;
+    alignas(64) std::atomic<VersionedIndex> _readCursor;
+
+    // Special Cell state for when queue has zero size
```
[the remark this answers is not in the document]
olofkallander: Too heavy comments. They can be more compact and still useful.

concurrency/MpmcQueue.h
olofkallander: The contention will be on readcursor and write cursor, those must be separated.
I assume it will invalidate the cacheline if it has been modified but I think that only is valid for atomic variables. If you read a variable that is not atomic like _capacity or _elements, you get whatever value is cached. CPU has not obligation to provide atomic access because you had an atomic in the cache line.
I think the cacheLineSeparator2 was intended to separate _writeCursor from atomics after the MpmcQueue. Otherwise that could cause contention with other unknown object containing atomics.

olofkallander (commented):
I would make the comments more concise. (really short and to the point). They take space and make it harder to read the real code.

concurrency/MpmcQueue.h
```diff
@@ -249,13 +282,21 @@ class MpmcQueue
...
+    const EntrySate _nullEntry = CellState::State::nullSlot;
+
+    alignas(64) std::atomic<VersionedIndex> _writeCursor;
+
+    // Although they are const. Cache can be invalidated for the reader when _writeCursor
+    // Perhaps is not a big issue (and even compile can optimize it this const values)
+    // but forcing a new cache line also ensures tha nothing is placed in this cache line after
+    // MpmcQueue boundary (which we can predict if can affect performance
+    // of _writeCursor)
+    alignas(64) const uint32_t _capacity;
```
olofkallander, line 299: Having thought this through, this one should not be aligned. It is not read/written atomically and the value will be from whatever is in cache and it will not stall cpu pipeline to read it (which is ok as it is const). 
It is important that writecursor is protected from atomic operations in the next cache line, but that should  be resolved by all other objects doing atomic operations align their first atomic member variable. Preferably having the atomic one first to not waste memory.

concurrency/MpmcHashmap.h
RicardoMDomingues: this was a mistake. Made it public only to print the size of entry in another place but it supposed to be private

concurrency/MpmcQueue.h
RicardoMDomingues: I remove his alignas. Then `_nullEntry`, `_capacity` and `_elements` are in the same cache line of _writeCursor to now waste space.

the `_writecursor` is protected from other atomic variables outside of this object because the `MpmcQueue` as an alignment requirement of 64 and the size needs to be a multiple of 64, so in this case the size is 128. Although there is no standardization on binary level, I think compilers don't have another option than forcing to be a multiple of aligment due to arrays (Example, if we have a MpmcQueue[N], accessing the memory position should `index * sizeof(MpmcQueue)` and that needs always be a 64 bytes aligned, then is ensured no other variables are added to the cache line outside of this object)

bridge/Mixer.cpp
```diff
@@ -289,10 +291,28 @@ bool Mixer::addBundleTransportIfNeeded(const std::string& endpointId, const ice:
         }
     }
 
+    const bool canDoVideo = hasVideoEnabled && !this->hasVideoDisabled();
+    const bool enableUplinkEstimation = canDoVideo;
+    const size_t expectedInboundStreamCount = canDoVideo ? 16 : 8;
+    const size_t expectedOutboundStreamCount = canDoVideo ? 256 : 16;
+    const size_t jobQueueSize = canDoVideo ? 4096 : 1024;
+
```
RicardoMDomingues, line 299: Better adjustment of transport memory allocated

api/AllocateConference.h
```diff
@@ -11,6 +11,7 @@ struct AllocateConference
 {
     utils::Optional<uint32_t> lastN;
     bool useGlobalPort = true;
+    bool disableVideo = false;
```
olofkallander, line 14: I prefer to avoid negated names. 
enableVideo = true;

bridge/Mixer.h
```diff
@@ -95,16 +95,19 @@ class Mixer
...
-        bool useGlobalPort,
-        VideoCodecSpec videoCodecs);
+        VideoCodecSpec videoCodecs,
+        bool useGlobalPort);
 
     virtual ~Mixer() = default;
 
     void markForDeletion();
     bool isMarkedForDeletion() const { return _markedForDeletion; }
+    bool hasVideoDisabled() const { return _videoSsrcs.empty(); }
```
olofkallander, line 105: hasVideoEnabled() const {return !_videoSsrcs.empty();}

test/bridge/MixerTest.cpp
```diff
@@ -51,11 +51,11 @@ const bridge::RtpMap FEEDBACK_RTP_MAP = bridge::RtpMap(bridge::RtpMap::Format::R
         if (!(EXPRESSION))                                                                                             \
         {                                                                                                              \
             FAIL() << (std::string()                                                                                   \
-                           .append("Line ")                                                                            \
-                           .append(std::to_string(__LINE__))                                                           \
-                           .append(": expression has failed!. Expression: '")                                          \
-                           .append(#EXPRESSION)                                                                        \
-                           .append("'"));                                                                              \
+                    .append("Line ")                                                                                   \
```
olofkallander, line 54: Is this due to new clang-format version? They changed something right?

bridge/engine/Engine.h
```diff
@@ -27,6 +27,7 @@ class Engine
 {
 public:
     Engine(jobmanager::JobManager& backgroundJobQueue);
+    Engine(jobmanager::JobManager& backgroundJobQueue, std::thread&& outsideThread);
```
olofkallander, line 30: externalThread is perhaps better name.

bridge/engine/Engine.cpp
```diff
@@ -35,6 +35,16 @@ Engine::Engine(jobmanager::JobManager& backgroundJobQueue)
...
 }
 
+Engine::Engine(jobmanager::JobManager& backgroundJobQueue, std::thread&& outsideThread)
+    : _messageListener(nullptr),
+      _running(true),
+      _tickCounter(0),
+      _tasks(1024),
+      _thread(std::move(outsideThread))
+{
+    // This construct is for unit test so it does not initialize a thread
```
olofkallander, line 45: Are there tests where this is used??

bridge/endpointActions/BarbellActions.cpp
```diff
@@ -102,7 +105,7 @@ httpd::Response generateBarbellResponse(ActionContext* context,
             addVp8VideoProperties(responseVideo);
         }
         addDefaultVideoProperties(responseVideo);
-        channelsDescription.video = responseVideo;
+        channelsDescription.video = std::move(responseVideo);
```
olofkallander, line 108: nice

olofkallander (approved):
Looks reasonable.
Only remark is I would suggest avoiding negated variable names like disabledB. It becomes harder to read boolean expressions. 
[Clean Code Prentice Hall]

test/bridge/MixerTest.cpp
RicardoMDomingues: not sure. Perhaps it was not properly formatted before

bridge/engine/Engine.cpp
RicardoMDomingues: Yes, My new unit tests on API level `ApiRequestHandlerTest.cpp` It starts a MixerManager with an `Engine` with an empty thread `std::thread`. It is a bit complex. So we start an `MixerManagerSpy` (which extend from `MixerManager` to have access to the `protected` fields) and that class intialize an `Engine` with an empty thread

https://github.com/finos/SymphonyMediaBridge/blob/96300d2e1fc5edd18907e7c2d29012e8ff69416b/test/include/mocks/MixerManagerSpy.h#L62

RicardoMDomingues:
> Looks reasonable. Only remark is I would suggest avoiding negated variable names like disabledB. It becomes harder to read boolean expressions. [Clean Code Prentice Hall]

Yes, I changed. I don't like much having default `true` values for boolean variables but that also is not a rule and we use default `true` in a lot of places. I agree for the readability is better not having the negative names

PR #405, opened 2025-04-15
MEDIA-3661: fix incorrect packet loss stats calculation

reddvl1980:


stepanaksenov (commented):
I'd suggest to update README also. Now it says:

> loss_download_hist is histogram for number of calls with certain average loss rate when sending to SMB. The buckets are [0,1,2,4,8, 100] percent.

https://github.com/finos/SymphonyMediaBridge/blob/master/doc/api/READMEapi.md#retrieve-metrics

In fact the buckets are

```
[0..1%)
[1..2%)
[2..4%)
[4..8%)
[8..16%)
[16.100%]
```

stepanaksenov (approved):

PR #409, opened 2026-04-07
MEDIA-4876: add support for bigger data channel message size

reddvl1980:
- Introdicing "PoolBuffer" - multi-chunk memory buffer to store sctp messages.
- Code is optimized so that when the message fit into single chunk copying on readong is not performed. 
- For temporary storage memory::Array<2048> is bing used so that if the message size fits 2048 bytes it's allocated on stack.
- Conference actions now return data['max-message-size'] - needed by SFU to return SDP to the c9 Trader Client. Max-message-size is omitted in barbell replys.
- Max message size is configured via config.sctp.maxMessageSize with the default value of 2048
- Unit tests added for PoolBuffer, sending pipeline, full forwarding pipline (requried adding of Engine mock).


bridge/engine/Engine.h
```diff
@@ -23,24 +23,24 @@ class MixerManagerAsync;
 struct EngineRecordingStream;
 struct RecordingDescription;
 
-class Engine
+class Engine : public IEngine
```
RicardoMDomingues: Small performance improvement. Avoid to read vtable when pointer points to an `Engine` object
 
```suggestion
class Engine final : public IEngine
```

bridge/engine/IEngine.h
```diff
@@ -0,0 +1,33 @@
...
+#include "utils/Function.h"
+
+namespace bridge
+{
+
+class MixerManagerAsync;
+class EngineMixer;
+struct RecordingDescription;
+
+class IEngine
```
RicardoMDomingues: `IEngine` is not consistent with the other names.

It should be `Engine` and `EngineImpl`

bridge/engine/EngineMixer.h
```diff
@@ -354,6 +356,7 @@ class EngineMixer : public transport::DataReceiver
     };
 
     using IncomingPacketInfo = IncomingPacketAggregate<memory::UniquePacket>;
+    using IncomingLargeSctpPacketInfo = IncomingPacketAggregate<memory::UniquePoolBuffer<memory::PacketPoolAllocator>>;
```
RicardoMDomingues: I don't think "Large" should be part of this name.
Perhaps `IncomingSctpPacketInfo` or `IncomingSctpMessageInfo`

memory/PoolBuffer.h
```diff
@@ -0,0 +1,348 @@
...
+};
+
+template <typename T>
+using UniquePoolBuffer = std::unique_ptr<PoolBuffer<T>>;
+
+template <typename TPoolAllocator>
+inline UniquePoolBuffer<TPoolAllocator> makeUniquePoolBuffer(TPoolAllocator& allocator,
+    size_t length)
+{
+    auto buffer = std::unique_ptr<PoolBuffer<TPoolAllocator>>(new PoolBuffer<TPoolAllocator>(allocator));
```
RicardoMDomingues: dynamic memory allocation. PoolBuffer needs to be allocated from the memory pool

memory/PoolBuffer.h
```diff
@@ -0,0 +1,348 @@
...
+
+    const std::vector<ChunkPtr>& getChunks() const { return _chunks; }
+
+    bool isNullTerminated() const {
+        return getReader().isNullTerminated();
+    }
+
+private:
+    TPoolAllocator& _allocator;
+    std::vector<ChunkPtr> _chunks;
```
RicardoMDomingues: Dynamic memory here as well.

I don't think the Buffer needs to be resizable. It defines a capacity from beginning and it should be allocate initialize the necessary headers to the pointers to the mem chunks based on the initial defined capacity. No memory should be allocated outside of the pool memory

memory/PoolBuffer.h
```diff
@@ -0,0 +1,348 @@
...
+    ReadonlyMemoryBuffer getReadonlyBuffer(ReadMode readMode = ReadMode::AsIs) const
+    {
+        ReadonlyMemoryBuffer result;
+        const bool nullTerminate = (readMode == ReadMode::NullTerminated);
+
+        if (_size == 0)
+        {
+            if (nullTerminate)
+            {
+                result.storage = std::make_unique<memory::Array<char, 1500>>(1);
```
RicardoMDomingues: can't allocate dynamic memory

memory/PoolBuffer.h
```diff
@@ -0,0 +1,348 @@
...
+
+        result.length = targetSize;
+
+        if (!needsCopy)
+        {
+            result.data = _chunks[0].get();
+        }
+        else
+        {
+            result.storage = std::make_unique<memory::Array<char, 1500>>(targetSize);
```
RicardoMDomingues: can't allocate dynamic memory

memory/PoolBuffer.h
```diff
@@ -0,0 +1,348 @@
...
+        {
+            targetSize++;
+            needsCopy = true;
+        }
+
+        result.length = targetSize;
+
+        if (!needsCopy)
+        {
+            result.data = _chunks[0].get();
```
RicardoMDomingues: lost track of memory ownership

memory/PoolBuffer.h
```diff
@@ -0,0 +1,348 @@
...
+            offsetInChunk = 0;
+            chunkIndex++;
+        }
+
+        return bytesWritten;
+    }
+
+    view getReader() const { return view(_chunks, _size, 0, _allocator); }
+
+    ReadonlyMemoryBuffer getReadonlyBuffer(ReadMode readMode = ReadMode::AsIs) const
```
RicardoMDomingues: in general I don't think we need this method at all. The buffer is not a continues memory and it does not need to be. In the end memory have to be copied to other place before be sent, we just need some helpers "copyTo" to make the memory copy to the final destination

memory/PoolBuffer.h
```diff
@@ -0,0 +1,348 @@
...
+
+#include "memory/PoolAllocator.h"
+#include "memory/Array.h"
+#include <vector>
+#include <memory>
+
+namespace memory
+{
+
+enum class ReadMode
```
RicardoMDomingues: Why we need this? I see the Buffer as a generic thing. outside context must now what is in the buffer. Buffer should be just a container of bytes that are not continuous stored in pool chunks. I would avoid making the buffer itself to holding contexts

transport/TransportImpl.cpp
```diff
@@ -185,43 +178,35 @@ class SctpSendJob : public jobmanager::CountedJob
...
-                header.payloadProtocol,
-                header.data(),
-                _packet->getLength() - sizeof(header),
+
+        const auto& continuousBuffer = _payload->getReadonlyBuffer();
+
+        if (!_sctpAssociation.sendMessage(_streamId,
+                _protocolId,
+                continuousBuffer.data,
+                continuousBuffer.length,
```
RicardoMDomingues: You had a unique buffer pointer! Avoid to convert it again in a raw pointers with a variable to hold the lenght. Just move the buffer until the memory is written to the final target

bridge/Mixer.cpp
```diff
@@ -1835,14 +1835,13 @@ void Mixer::sendEndpointMessage(const std::string& toEndpointId,
     const utils::SimpleJson& message)
 {
     assert(fromEndpointIdHash);
-    if (message.size() >= memory::AudioPacket::maxLength())
+    if (message.size() >= _config.sctp.maxMessageSize)
     {
         return;
```
RicardoMDomingues, line 1844: Perhaps add a warn log here to this not be dropped silently

bridge/engine/EngineMixer.cpp
```diff
@@ -586,38 +605,71 @@ void EngineMixer::sendEndpointMessage(const size_t toEndpointIdHash,
             return;
         }
 
-        api::DataChannelMessage::makeEndpointMessage(endpointMessage,
-            toDataStream->endpointId,
-            fromDataStream->endpointId,
+        int length = 0;
+#if ENABLE_LEGACY_API
```
RicardoMDomingues: Why do we need to transport all this complexity to `EngineMixer`? Can't be the `makeEndpointMessage` changed to produce a PollBuffer with final message?

bridge/engine/EngineMixerBarbell.cpp
```diff
@@ -511,35 +512,58 @@ SsrcInboundContext* EngineMixer::emplaceBarbellInboundSsrcContext(const uint32_t
...
+            const char* message = reinterpret_cast<const char*>(sctpHeader->data());
+            const auto messageLength = continuousBuffer.length - sizeof(*sctpHeader);
             if (messageLength == 0)
             {
-                return;
+                continue;
             }
 
             auto messageJson = utils::SimpleJson::create(message, messageLength);
+            std::string messageStr(message, messageLength);
```
RicardoMDomingues: Dynamic memory is not allowed on engine. Here you have to use the StringBuilder and copy the memory to the stack OR adjust the JsonReader to use "LegacyRandomAccessIterator" interfaces and implement a LegacyRandomAccessIterator for the PoolBuffer. In that way the json could be parsed even without make it contigous. While I really think the PoolBuffer should implement the std container interface with a begin() and end() that returns a LegacyRandomAccessIterator, moving JsonReader to use generic iterators perhaps it would be a more complex task, then is it ok we just copy this to a contiguous memory in the stack

bridge/engine/Engine.h
reddvl1980: I removed IEngine, EngineMock and related changes and just added miminal mock in the test.

bridge/engine/EngineMixer.h
reddvl1980: Renamed to IncomingSctpMessageInfo.

bridge/Mixer.cpp
reddvl1980: done

memory/PoolBuffer.h
reddvl1980: fixed

bridge/engine/EngineMixerBarbell.cpp
reddvl1980: fixed

memory/PoolBuffer.h
reddvl1980: Fixed.

memory/PoolBuffer.h
reddvl1980: agree

memory/PoolBuffer.h
reddvl1980: removed

transport/TransportImpl.cpp
reddvl1980: agreed

bridge/engine/EngineMixer.cpp
reddvl1980: done

memory/PoolBuffer.h
reddvl1980: fixed

memory/PoolBuffer.h
reddvl1980: fixed, but pease check

legacyapi/DataChannelMessage.h
```diff
@@ -25,6 +27,40 @@ inline void makeEndpointMessage(utils::StringBuilder<2048>& outMessage,
...
+    }
+
+    auto written = buffer->copyFrom(TO_STRING, std::char_traits<char>::length(TO_STRING), 0);
+    written += buffer->copyFrom(toEndpointId.c_str(), toEndpointId.length(), written);
+    written += buffer->copyFrom(FROM_STRING, std::char_traits<char>::length(FROM_STRING), written);
+    written += buffer->copyFrom(fromEndpointId.c_str(), fromEndpointId.length(), written);
+    written += buffer->copyFrom(MSG_STRING, std::char_traits<char>::length(MSG_STRING), written);
+    written += buffer->copyFrom(*payload.get(), written);
+    written += buffer->copyFrom(TAIL_STRING, std::char_traits<char>::length(TAIL_STRING), written);
+
```
RicardoMDomingues: You have calculated all that lengths before. Perhaps you could stores it in a variable to avoid calculate them again

memory/PoolBuffer.h
```diff
@@ -0,0 +1,493 @@
...
+        explicit Deleter(TPoolAllocator& allocator) : _allocator(&allocator) {}
+
+        void operator()(PoolBuffer<TPoolAllocator>* p)
+        {
+            if (p)
+            {
+                p->~PoolBuffer();
+                if (_allocator)
+                {
+                    _allocator->free(p);
```
RicardoMDomingues, line 37: You can not do this when your move the data as the allocated memory now belongs to other objects.

I think there is 3 options.
 - You don't do this when you lose the ownership. Only release when when PoolBuffer that has the ownership is deleted
 - Copy always the content master chunk on move constructor/oerator
 - Perhaps don't allocate PoolBuffer dynaically (use it like a unique_ptr itslef, where you allow move internal dynamic memory but the contol data is always in the stack [I thin this option is preferable]

legacyapi/DataChannelMessage.h
reddvl1980: We use c++17, and starting in C++17, `std::char_traits<char>::length` is officially declared as constexpr.
The argument , e.g. TAIL_STRING, and otehrs, is explicitly declared as a constexpr variable.

memory/PoolBuffer.h
reddvl1980: Fixed.
I've also updated header files to clarify semantics of chaning ownership (added && and const& where needed).

memory/PoolBuffer.h
```diff
@@ -0,0 +1,418 @@
...
+#include "memory/PoolAllocator.h"
+#include "memory/PacketPoolAllocator.h"
+#include "memory/Array.h"
+#include <vector>
+#include <memory>
+
+namespace memory
+{
+//template<size_t SIZE>
+struct ReadonlyMemoryBuffer
```
RicardoMDomingues: leftover?

memory/PoolBuffer.h
```diff
@@ -0,0 +1,418 @@
...
+            return true;
+        };
+
+        forEachBlock(sourceOffset, count, callback);
+        return bytesCopied;
+    }
+
+    TPoolAllocator& getAllocator() const { return _allocator; }
+
+    bool isNullTerminated() const {
```
RicardoMDomingues: I thing this method shouldn't live inside a generic container. Also don't find any usage of in the code so I guess it is dead code and can be removed?

memory/PoolBuffer.h
```diff
@@ -0,0 +1,418 @@
...
+                {
+                    numChunks = 1;
+                }
+
+                pointersAreaSize = numChunks * sizeof(void*);
+                if (masterChunkCapacity > pointersAreaSize)
+                {
+                    _firstChunkSize = masterChunkCapacity - pointersAreaSize;
+                    _numChunks = numChunks;
+                    _firstChunkIsInMaster = true;
```
RicardoMDomingues: code a little confusing and I don't know why do we need this flag at all. it feels like `_firstChunkIsInMaster` ==  _numChunks > 1. So its redundant. Code is a little confusing. Can we simplify it?

like:

```
const size_t chunkSize = _allocator.getElementSize();
size_t firstChunkDataOffset = 0;
size_t numAdditionalChunks = 0;

size_t availableSpace  = (chunkSize - firstChunkDataOffset) + numAdditionalChunks * chunkSize;

while (availableSpace < size) {
   size_t additionalSpaceNeeded = size - availableSpace;
   size_t newAdditionaChunks = (additionalSpaceNeeded + elementSize - 1)  / chunkSize;
 
  firstChunkDataOffset += sizeof(void*) * newAdditionaChunks;
  numAdditionalChunks + = newAdditionaChunks
}

if (firstChunkDataOffset > chunkSize + 1) {

    logger::error("master chunk is too small to hold chunk pointers. capacity %zu, required %zu",
                    "PoolBuffer",
                    chunkSize,
                    _numChunks * sizeof(void*));

     return false;
}

_masterChunk = _allocator.allocate();
std::fill(_masterChunk, _masterChunk + firstChunkDataOffset, 0); // Set zero all pointers

void* additionalChunkPointer = _masterChunk[0];

for (size_t i = 0; i < numAdditionalChunks; i++) {
 additionalChunkPointer = _allocator.allocate();
 if (!additionalChunkPointer) {
      // Release memeory
  }
}

...
```

memory/PoolBuffer.h
```diff
@@ -0,0 +1,418 @@
...
+                    _numChunks = 0;
+                    return false;
+                }
+            }
+        }
+        _size = size;
+        return true;
+    }
+
+    void clear()
```
RicardoMDomingues: Not sure if this method should be call "clear". On stl containers, clear don't release memory. That means after the "clean" and new "allocation" needs to be called.

Perhaps rename it as "deallocate" or something and make private?

memory/PoolBuffer.h
reddvl1980: removed!

memory/PoolBuffer.h
reddvl1980: Yes, it's a dead code. Removed.

memory/PoolBuffer.h
reddvl1980: I guess the best way would be 'reset', it's usually used for smart pointers, by deallocating storage, and resetting it to the 'null' state. It's not exactly that, but it's better than 'clear' that is supposed to keep capacity.

memory/PoolBuffer.h
```diff
@@ -0,0 +1,336 @@
...
+    PoolBuffer& operator=(PoolBuffer&& other) noexcept
+    {
+        if (this != &other)
+        {
+            free();
+            _masterChunk = other._masterChunk;
+            _size = other._size;
+            _numAdditionalChunks = other._numAdditionalChunks;
+            _firstChunkDataOffset = other._firstChunkDataOffset;
+            other._masterChunk = nullptr;
```
RicardoMDomingues: for correctness you should also store the allocator (you have to change to a pointer first). While it is required to be the same template type, it is not required to be the same object. So we have to store the allocator were the chunks were allocated

RicardoMDomingues (approved):

memory/PoolBuffer.h
reddvl1980: Fixed and optmimized further.

memory/PoolBuffer.h
reddvl1980: I've opted to delete move= operator altogether as it is not used in the prod code, only in single test (also removed). Alternative would be to keep operator but switch storage of _allocator from & to *.

PR #416, opened 2026-08-07
Add production Docker image and compose setup

marquies:
Multi-stage Dockerfile that reproduces the docker/ubuntu-jammy toolchain (clang/llvm 12 + libc++, OpenSSL 3.4, libsrtp 2.6, libmicrohttpd, opus) in a builder stage and ships a slim ubuntu:jammy runtime carrying only the Release `smb` binary and its runtime shared libraries.

The per-distro folders under docker/ are build/CI images; this adds a deployable runtime image alongside them.

- docker-compose.yml defaults to host networking (SFU advertises its own reachable IP in ICE candidates and carries media over UDP), with a bridged published-ports alternative documented inline.
- docker/runtime-config.json binds the HTTP API to 0.0.0.0 and collapses media onto a single shared ICE port to keep the port surface small.
- Runs non-root by default; CAP_SYS_NICE for SCHED_FIFO is opt-in and deliberately not baked on with setcap, since permitted caps outside the container bounding set make the binary fail to exec at all.
- docker/CONTAINER.md documents build, both network topologies, the config keys, and the real-time priority requirements.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,31 @@
+# Keep the build context small and reproducible.
+# NOTE: .git is intentionally NOT ignored — version/CheckGit.cmake needs it
+# to stamp git_version.h during the build.
+
+# In-source CMake artifacts (README documents running `cmake .` in-tree)
+CMakeCache.txt
+CMakeFiles/
+cmake_install.cmake
+Makefile
+build/
+out/
+*.o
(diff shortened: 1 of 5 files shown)


marquies:
Wrong base repo, sorry for the inconvinience.

